PR #6, opened 2018-04-09
Fix bug of mapping dictating

t-liyari:


azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
```diff
@@ -95,3 +95,10 @@ def __init__(self, database, table,
         self.report_level = reportLevel
         self.report_method = reportMethod
         self.validation_policy = validationPolicy
+
+    def get_mapping_format(self):
```
vladikbr: Please add support for avro

PR #18, opened 2018-06-04
AAD Helper bug fix: in version 1.0.0 authorily url should not end with /

toshetah:


stcheng:
https://github.com/AzureAD/azure-activedirectory-library-for-python/pull/157

actually I have talked with ADAL team and they relax the check for this URL

PR #41, opened 2018-07-31
Edit data ingest sample

toshetah:


azure-kusto-ingest/tests/sample.py
```diff
@@ -2,8 +2,8 @@
 
 ingestion_properties = IngestionProperties(database="database name", table="table name", dataFormat=DataFormat.csv)
 
-ingest_client = KustoIngestClient("https://ingest-clustername.kusto.windows.net")
-ingest_client = KustoIngestClient("https://ingest-clustername.kusto.windows.net", client_id="aad app id", client_secret="secret")
+ingest_client = KustoIngestClient("https://ingest-<clustername>.kusto.windows.net")
+ingest_client = KustoIngestClient("https://ingest-<clustername>.kusto.windows.net", client_id="aad app id", client_secret="secret")
 
 file_descriptor = FileDescriptor("E:\\filePath.csv", 3333) # 3333 is the raw size of the data in bytes.
```
vladikbr: Change this to something that won't compile? Like: 
FileDescriptor("<local file path>", uncompressedSizeBytes) # specify uncompressed file size in bytes.

PR #48, opened 2018-08-08
Make better iterables

toshetah:
Make 3 iterables:
_KustoResponseDataSet
_KustoResultTable
_KustoResultRow

In addition: Change private modules to start with _


azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +192,116 @@ def _get_table(self, table_id):
...
+            self.tables_count = len(self.tables)
+
+    @property
+    def primary_results(self):
+        if self.tables_count == 1:
+            return self.tables[0]
+        primary = list(filter(lambda x: x.table_kind == "PrimaryResult", self.tables))
+        if len(primary) == 1:
+            return primary[0]
+        return primary
```
t-liyari: It can be more than one table with "PrimaryResult" as table_kind ?
Don't quite understand the separation of when len(primary) equals 1 and when not.

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -93,83 +94,61 @@ def to_timedelta(value):
...
-        """ Gets the excpetions got from Kusto if exists. """
-        if self.has_exceptions():
-            if isinstance(self.json_response, list):
-                return list(
-                    filter(lambda x: x["FrameType"] == "DataSetCompletion", self.json_response)
-                )[0]["OneApiErrors"]
-            return self.json_response["Exceptions"]
-        return None
+    def __getitem__(self, key):
+        return _KustoResultRow(self.columns_count, self.columns, self._rows[key])
```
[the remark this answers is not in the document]
toshetah: We do not support at the moment getting a column. Only complete lines.

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +192,116 @@ def _get_table(self, table_id):
...
+
+class _DataSetCompletion(object):
+    """Completion of response. Exists only in v2(queries)"""
+
+    def __init__(self, json):
+        self.has_errors = json["HasErrors"]
+        self.cancelled = json["Cancelled"]
+        self.errors = json["OneApiErrors"] if "OneApiErrors" in json else None
+
+
```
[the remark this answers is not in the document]
toshetah: I thought about it. Not yet. :)

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +192,116 @@ def _get_table(self, table_id):
...
+class _KustoResponseDataSet(six.Iterator):
+    """Wrapper for response."""
+
+    def __init__(self, json_response, is_v2):
+        self.is_v2 = is_v2
+        self._index = 0
+        if self.is_v2:
+            deta_set_header = list(
+                filter(lambda x: x["FrameType"] == "DataSetHeader", json_response)
+            )[0]
```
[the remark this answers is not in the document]
toshetah: This code is not there anymore

azure-kusto-data/azure/kusto/data/_kusto_client.py
toshetah: Yes. Think about the following query:
T | fork
( take 1 )
( take 1 )

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +202,135 @@ def _get_table(self, table_id):
...
+                result.append(
+                    "Please provide the following data ot Kusto: CRID='{0}' Description:'{1}'".format(
+                        query_status_table[i][self._crid_column],
+                        query_status_table[i][self._status_column],
+                    )
+                )
+        return result
+
+    def __iter__(self):
+        return self
```
zooba: Looks like you could simply return `iter(self.tables)` here and then this doesn't have to be an iterator?

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +202,135 @@ def _get_table(self, table_id):
...
+            self.tables_count = len(json_response["Tables"])
+        else:
+            json_tables = json_response["Tables"][:-1]
+            toc = _KustoResultTable(json_response["Tables"][-1])
+            for i in range(0, len(json_tables)):
+                json_table = json_tables[i]
+                json_table["TableName"] = toc[i]["Name"]
+                json_table["TableId"] = toc[i]["Id"]
+                json_table["TableKind"] = self._v1_tables_kinds[toc[i]["Kind"]]
+                self.tables.append(_KustoResultTable(json_table))
```
zooba: If I end up in this path while directly constructing `_KustoResponseDataSetV1` then `self.tables` doesn't exist here.

You probably want to make this hierarchy work with the `super` call being the first thing in each initializer. That'll require a bit of refactoring.

(Even better if you only have one initializer on the base class, and it implicitly handles the different versions by simply overriding the class members. But I'm not sure you can do that trivially.)

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +202,135 @@ def _get_table(self, table_id):
...
+class _KustoResponseDataSetV2(_KustoResponseDataSet):
+
+    _status_column = "Payload"
+    _error_column = "Level"
+    _crid_column = "ClientRequestId"
+
+    def __init__(self, json_response):
+        json_response = [t for t in json_response if t["FrameType"] == "DataTable"]
+        for json_table in json_response:
+            json_table["TableKind"] = WellKnownDataSet[json_table["TableKind"]]
```
zooba: Intentionally modifying the contents of an argument?

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +202,135 @@ def _get_table(self, table_id):
...
+    def get_exceptions(self):
+        """Gets the excpetions retrieved from Kusto if exists."""
+        query_status_table = next(
+            (t for t in self.tables if t.table_kind == WellKnownDataSet.QueryCompletionInformation),
+            None,
+        )
+        if not query_status_table:
+            return
+        result = []
+        for i in range(0, len(query_status_table)):
```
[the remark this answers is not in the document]
toshetah: Fixed.

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -223,6 +202,135 @@ def _get_table(self, table_id):
...
+        if self._index >= self.tables_count:
+            raise StopIteration
+        result = self.tables[self._index]
+        self._index += 1
+        return result
+
+    def __getitem__(self, key):
+        if isinstance(key, numbers.Number):
+            return self.tables[key]
+        return self.tables[self._tables_names.index(key)]
```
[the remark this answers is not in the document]
toshetah: Fixed.

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -93,83 +104,61 @@ def to_timedelta(value):
...
+class _KustoResultColumn(object):
+    def __init__(self, json_column, ordianl):
+        self.column_name = json_column["ColumnName"]
+        self.column_type = (
+            json_column["ColumnType"] if "ColumnType" in json_column else json_column["DataType"]
+        )
+        self.ordinal = ordianl
+
+
+class _KustoResultTable(six.Iterator):
```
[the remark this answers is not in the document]
toshetah: Fixed.

azure-kusto-data/azure/kusto/data/_kusto_client.py
toshetah: Fixed.

azure-kusto-data/azure/kusto/data/_kusto_client.py
toshetah: Yes. I want the table kind to be from the defined enum. The response us just text.

azure-kusto-data/azure/kusto/data/_kusto_client.py
```diff
@@ -285,50 +270,40 @@ def to_dataframe(self):
...
-        for json_table in json_response:
-            json_table["TableKind"] = WellKnownDataSet[json_table["TableKind"]]
-        self.tables = [_KustoResultTable(json_table) for json_table in json_response]
-        super(_KustoResponseDataSetV2, self).__init__()
+        super(_KustoResponseDataSetV2, self).__init__(
+            [t for t in json_response if t["FrameType"] == "DataTable"]
+        )
+        self._status_column = "Payload"
+        self._error_column = "Level"
+        self._crid_column = "ClientRequestId"
```
zooba: These assignments may not be necessary - the class level variables you had were fine. (Even during `__init__`, as base classes are initialized they will see the most-derived class variables when they look at `self.<name>`. It's only if they do `_KustoResponseDataSetV1.<name>` that they'll get the exact value on that type.)

azure-kusto-data/tests/test_functional.py
```diff
@@ -229,13 +229,12 @@ def test_column_dont_exist(self):
...
         """Tests StopIteration is raised when the response ends."""
         response = _KustoResponseDataSetV2(json.loads(RESPONSE_TEXT))
-        iterator = response.primary_results
-        iterator.__next__()
-        iterator.__next__()
-        iterator.__next__()
-        self.assertRaises(StopIteration, iterator.__next__)
+        row_count = 0
+        for _ in response.primary_results:
+            row_count += 1
```
zooba: Some alternatives you could use here if you want:

```python
row_count = len(list(response.primary_results))
row_count = sum(1 for _ in response.primary_results)
```

toshetah: Fixed.

PR #60, opened 2018-08-24
Determine default blob size

jmandel:
It would be nice not to require a size argument for blob descriptors -- or else to have the client auto-compute one if the caller leaves it out. The equivalent of the attached...


danield137:
Closing due to staleness.

PR #64, opened 2018-09-13
Ingestion status

danield137:


azure-kusto-ingest/tests/sample.py
```diff
@@ -1,21 +1,55 @@
...
 from azure.kusto.data.request import KustoConnectionStringBuilder
 from azure.kusto.ingest import KustoIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
 
-INGESTION_PROPERTIES = IngestionProperties(database="database name", table="table name", dataFormat=DataFormat.csv)
+ingestion_props = IngestionProperties(database="{database_name}", table="{table_name}", dataFormat=DataFormat.csv)
+client = KustoIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
 
-INGEST_CLIENT = KustoIngestClient("https://ingest-<clustername>.kusto.windows.net")
+# there are more posbilities for authenticating
+# KCSB = KustoConnectionStringBuilder.with_aad_application_key_authentication(
```
vladikbr: Maybe show here the use of authority ID?

danield137: this is just an older branch, once I will rebase I will make sure that there would be such an example

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -1,5 +1,3 @@
-"""Kusto ingest client for Python."""
```
toshetah: Doesn't it cause pylint to tell you that you need to write a module dicstring?

azure-kusto-ingest/tests/e2e.py
```diff
@@ -39,7 +42,7 @@ def create_deft_table_csv_mappings():
         mappings.append(CsvColumnMapping(columnName="xint64", cslDataType="long", ordinal=7))
         mappings.append(CsvColumnMapping(columnName="xuint8", cslDataType="long", ordinal=8))
         mappings.append(CsvColumnMapping(columnName="xuint16", cslDataType="long", ordinal=9))
-        mappings.append(CsvColumnMapping(columnName="xuint32", cslDataType="long", ordinal=10))
+        mappings.append(CsvColumnMapping(columnName="xuint32", cslDataType="long", ordinal=20))
```
toshetah: Please undo to 10.

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
danield137: It does actually, but since it wasn't very helpful I removed it. I can make sure all files have them . not sure they have any value (as classes and functions are documented).

Other azure projects tend to just put license information instead...

azure-kusto-ingest/azure/kusto/ingest/_status_q.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            for queue_details in self.get_queues_func()
+        ]
+
+    def is_empty(self):
+        """Checks if Status queue has any messages        
+        """
+        return not self.peek(1, raw=True)
+
+    def decode_content(self, content):
```
toshetah: underscore?

azure-kusto-ingest/azure/kusto/ingest/_status_q.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # TODO: current implementation takes a union top n /  len(queues), which is not ideal,
+    # because the user is not supposed to know that there can be multiple underlying queues
+    def pop(self, n=1, raw=False, delete=True):
+        """Pop status queue
+        :param int n: number of messages to return as part of peek.
+        :param bool raw: should message content be returned as is (no parsing).
+        :param bool delete: should message be deleted after pop. default is True as this is expected of a q.
+        """
+        q_services = self._get_q_services()
```
toshetah: Random

PR #65, opened 2018-09-13
allow passing authority id to KustoConnectionStringBuilder

danield137:


azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -62,6 +62,9 @@ def __init__(self, connection_string):
         self._internal_dict = {}
         if connection_string is not None and "=" not in connection_string.partition(";")[0]:
             connection_string = "Data Source=" + connection_string
+
+        self[self.ValidKeywords.authority_id.value] = "common"
```
toshetah: Please remove the .value

PR #74, opened 2018-11-18
Add anonymous authentication

toshetah:


azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -107,6 +111,7 @@ def with_aad_application_key_authentication(cls, connection_string, aad_app_id,
         _assert_value_is_valid(app_key)
         _assert_value_is_valid(authority_id)
         kcsb = cls(connection_string)
+        kcsb[kcsb.ValidKeywords.aad_federated_security] = True
```
danield137: why do you assume it's federated? 
what I could find here https://docs.microsoft.com/en-us/azure/active-directory/hybrid/how-to-connect-fed-whatis explains that federation means it's on-premise active directory vs. the azure active directory...

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -232,7 +244,7 @@ def __init__(self, kcsb):
         kusto_cluster = kcsb.data_source
         self._mgmt_endpoint = "{0}/v1/rest/mgmt".format(kusto_cluster)
         self._query_endpoint = "{0}/v2/rest/query".format(kusto_cluster)
-        self._aad_helper = _AadHelper(kcsb)
+        self._aad_helper = _AadHelper(kcsb) if kcsb.aad_federated_security else None
```
danield137: following previous comment, this should be reversed. if federated, then no point of using aad.

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -289,17 +301,17 @@ def _execute(
...
-            "Authorization": access_token,
             "Accept": "application/json",
             "Accept-Encoding": "gzip,deflate",
             "Content-Type": "application/json; charset=utf-8",
-            "Fed": "True",
             "x-ms-client-version": "Kusto.Python.Client:" + VERSION,
             "x-ms-client-request-id": "KPC.execute;" + str(uuid.uuid4()),
         }
 
+        if self._aad_helper:
```
danield137: this is awkward and hard to read the intent from. why wouldn't I add auth if I have no aad_helper? the actual reasin I don't use auth, is because no `auth_provider` or `auth_method` was set, which is what I would expect to see here.

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -49,6 +50,8 @@ def parse(cls, key):
                 return cls.application_certificate_thumbprint
             if key in ["authority id", "authorityid", "authority", "tenantid", "tenant", "tid"]:
                 return cls.authority_id
+            if key in ["aad federated security", "federated security", "federated", "fed", "aadfed"]:
+                return cls.aad_federated_security
             raise KeyError(key)
 
     def __init__(self, connection_string):
```
danield137: I think we should explicitly expose `**kwargs` here and allow a `auth=False` to be set. would make code less cumbersome IMHO

azure-kusto-data/azure/kusto/data/request.py
vladikbr: I agree. Can we control this behavior by auth_mehtod being not set or equal to "Anonymous"?

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -289,17 +301,17 @@ def _execute(
...
             "Accept": "application/json",
             "Accept-Encoding": "gzip,deflate",
             "Content-Type": "application/json; charset=utf-8",
-            "Fed": "True",
             "x-ms-client-version": "Kusto.Python.Client:" + VERSION,
             "x-ms-client-request-id": "KPC.execute;" + str(uuid.uuid4()),
         }
 
+        if self._aad_helper:
+            request_headers["Authorization"] = self._aad_helper.acquire_token()
```
vladikbr: Maybe explicit call and error handling would be better?

azure-kusto-data/azure/kusto/data/request.py
vladikbr: Not quite. It is a common use to use "AAD Federated" as a denotion that we authenticate to one of the AAD tenants.

azure-kusto-data/tests/sample.py
```diff
@@ -31,9 +31,16 @@
...
+# Please note that if you choose this option, you'll need to autenticate for every new instance that is initialized.
+# It is highly recommended to create one instance and use it for all of your queries.
+kcsb = KustoConnectionStringBuilder.with_aad_device_authentication(cluster)
+
+
+# The authentication method will be taken from the chosen KustoConnectionStringBuilder.
 client = KustoClient(kcsb)
 
-# In case you want to authenticate with the logged in AAD user.
+# In case you do not want authentication. For instance if you run Kusto locally.
```
vladikbr: Please rephrase to "anonymous" instead of "no authentication"

azure-kusto-data/azure/kusto/data/request.py
toshetah: https://docs.microsoft.com/en-us/azure/kusto/api/connection-strings/kusto#application-authentication-properties

azure-kusto-data/azure/kusto/data/request.py
toshetah: Discussed F2F, we are aligned in out connection strings in all languages.

PR #81, opened 2018-12-10
Add ClientRequestProperties class

toshetah:


azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -312,19 +303,55 @@ def _execute(
...
-            return kusto_response
+                return KustoResponseDataSetV1(response.json())
         else:
             raise KustoServiceError([response.json()], response)
+
+    def _get_timeout(self, properties, default):
+        if properties:
+            if properties.has_option(ClientRequestProperties.OptionServerTimeout):
+                return properties.get_option_value_or_default(ClientRequestProperties.OptionServerTimeout, default)
+            return None
```
danield137: I don't think this is what you meant.. if `properties` exists, but not `has_option`, we should still revert to default.

I would just change the entire function to 

`properties.get_value(ClientRequestProperties.OptionServerTimeout,default) if properties else default`

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -312,19 +303,55 @@ def _execute(
...
+    def set_option(self, name, value):
+        """Sets an option's value"""
+        _assert_value_is_valid(name)
+        self.options[name] = value
+
+    def has_option(self, name):
+        """Checks if an option is specified."""
+        return name in self.options
+
+    def get_option_value_or_default(self, name, default_value):
```
danield137: very verbose for python, would change to `get_option`

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -312,19 +303,55 @@ def _execute(
...
+    """This class is a POD used by client making requests to describe specific needs from the service executing the requests.
+    For more information please look at: https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties
+    Not all of the documented options are implemented. You are welcome to open an issue in case you need one of them.
+    """
+
+    OptionDeferPartialQueryFailures = "deferpartialqueryfailures"  # TODO: Rename: results_defer_partial_query_failures
+    OptionServerTimeout = "servertimeout"  # TODO: Rename: request_timeout
+
+    def __init__(self):
+        self.options = {}
```
danield137: should probably be prefixed with `_` as we want to prevent access. otherwise class is kinda redundant

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -312,19 +303,55 @@ def _execute(
...
+        _assert_value_is_valid(name)
+        self.options[name] = value
+
+    def has_option(self, name):
+        """Checks if an option is specified."""
+        return name in self.options
+
+    def get_option_value_or_default(self, name, default_value):
+        """Gets an option's value."""
+        if self.has_option(name):
```
danield137: don't need to check -> `self.options.get(name,default)`

PR #96, opened 2019-01-22
Bug fix: convert timespan to pandas timedelta

toshetah:
Make a long story short:
Python datetime/timedelta(timespan) native types supports 6 decimal digits.
Kusto supports 7.
Pandas types supports 9.

Why am I telling this? We noticed that we don't give the existing accuracy to users that consumes pandas.
In order to do that we are keeping the 7th digit in the KustoResultRow class, and when converting into pandas we calculate back the 7th digit.
While doing that, we are now creating KustoResultRows on creation of the response, and not lazily.
The advantage in that is that one will receive True for the following:
r.primary_results[0][0] == r.primary_results[0][0]
In contrast to what we had so far.


azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -19,51 +21,94 @@ class WellKnownDataSet(Enum):
...
+            try:
+                lower_column_type = column.column_type.lower()
+            except AttributeError:
+                self._value_by_index.append(value)
+                self._value_by_name[columns[i]] = value
+                continue
+
+            if lower_column_type == "dynamic" and not value:
+                typed_value = json.loads("null")
+                # TODO: After Yifats change this if should be removed.
```
danield137: what change is that? can you elaborate

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -19,51 +21,94 @@ class WellKnownDataSet(Enum):
...
+                # The servers should not return empty string anymore as a valid json.
+            elif lower_column_type in ["datetime", "timespan"]:
+                if value is None:
+                    typed_value = None
+                else:
+                    try:
+                        # If you are here to read this, you probably hit some datetime/timedelta inconsistencies.
+                        # Azure-Data-Explorer(Kusto) supports 7 decimal digits, while the corresponding python types supports only 6.
+                        # What we do here, is removing the 7th digit, if exists, and creatign datetime/timedelta
```
danield137: english

```suggestion
                        # What we do here, is remove the 7th digit, if exists, and create a datetime/timedelta
```

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -19,51 +21,94 @@ class WellKnownDataSet(Enum):
...
+                        # from the remaining. The reason we are keeping the 7th digit, is to allow users getting
```
danield137:
```suggestion
                        # from whats left. The reason we are keeping the 7th digit, is to allow users to work with
```

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -19,51 +21,94 @@ class WellKnownDataSet(Enum):
...
+                        # this percision in case they want it. One example one might want this percision, is when
```
danield137:
```suggestion
                        # this precision in case they want it. One example why one might want this precision, is when
```

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -19,51 +21,94 @@ class WellKnownDataSet(Enum):
...
+                        # working with pandas. In that case, use azure.kusto.data.helpers.dataframe_from_result_table
+                        # which take into account the 7th digit.
```
danield137:
```suggestion
                        # which takes into account the 7th digit.
```

PR #102, opened 2019-01-30
Added connection pool capability

yaniv-ms:
Database connection pool reuses connections which are already open in a thread-safe way and therefore avoids TCP and TLS handshake for each query.
requests.post has been replaced with urllib3.PoolManager.request which manages the connection pool to a single server.
Requirement for requests>=2.13.0 has been replaced with urllib3[secure]>=1.20. There was a significant improvement in urllib 1.20 so it is the earliest version which has to be demanded. The [secure] flavor adds dependency in certifi which is used for TLS certification validation.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/request.py
+++ b/azure-kusto-data/azure/kusto/data/request.py
@@ -2,9 +2,11 @@
 
 import uuid
 import json
+import urllib3
+import certifi
+
 from datetime import timedelta
 from enum import Enum, unique
-import requests
 
 from .security import _AadHelper
 from .exceptions import KustoServiceError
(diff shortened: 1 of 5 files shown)


azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -240,14 +242,22 @@ class KustoClient(object):
     _mgmt_default_timeout = timedelta(hours=1, seconds=30).seconds
     _query_default_timeout = timedelta(minutes=4, seconds=30).seconds
 
-    def __init__(self, kcsb):
+    def __init__(self, kcsb, max_pool_size=100):
```
toshetah: Let's put max_pool_size as a static member.
This property is for advanced usage, and should not be part of the API.

PR #104, opened 2019-02-05
Update _models.py

miflower:
Add additional case for properly handling returned dynamic data

When data is stored in a Kusto table as a dynamic type and is formatted properly then it is parsed by the response handler as a JSON object and not as a string object.
Checking this value for the column and checking if it is already a dynamic type prevents a error being thrown by the json module for passing a dict type to loads.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_models.py
+++ b/azure-kusto-data/azure/kusto/data/_models.py
@@ -48,6 +48,8 @@ class KustoResultRow(object):
                 # There is current work being done to stay with the last, as this is the official representation for
                 # empty json. Once this work is done this if should be removed.
                 # The servers should not return empty string anymore as a valid json.
+            if lower_column_type == "dynamic" and isinstance(value, dict):
+                type_value = value
             elif lower_column_type in ["datetime", "timespan"]:
                 if value is None:
                     typed_value = None


toshetah:
Hello Miflower,

First of all, thanks for contributing!
This is handled at [line 80](https://github.com/Azure/azure-kusto-python/blob/6e97a4e1a0be237b861844ad1a8a124f2ab736f6/azure-kusto-data/azure/kusto/data/_models.py#L80).
The response is string, and should be loaded by json. Do you see any regression in json values?

miflower:
Hey Tomer,

The issue that I'm running into is that some of my underlying datasets are stored as dynamic type but aren't proper JSON objects.

Without the patch, using the below code:

```
res = kusto_ctx.query("""
datatable(EmptyColumn: string) [ "empty" ]
| extend GoodJSON=todynamic('{"key":"value"}')
""")

display(res)

type(res['GoodJSON'][0])
```

I get the table to print properly along with the type of the value being dict (as expected)

Using the below code:

```
res = kusto_ctx.query("""
datatable(EmptyColumn: string) [ "empty" ]
| extend GoodJSON=todynamic('ThisIsNotValidJSON')
""")

display(res)

type(res['GoodJSON'][0])
```

The library throws an exception on line 80 with "JSONDecodeError: Expecting value: line 1 column 1 (char 0)"

Hopefully this help to clear up the issue I'm facing, let me know if you need better dumps.

toshetah:
Thank you for your contribution Miflower.
We fixed that issue in [this commit](https://github.com/Azure/azure-kusto-python/commit/2aa575f1442c307ff5d7e4f021aa9d1760f1cb8b).

toshetah:
Can you please try version 0.0.23 and verify your issue is resolved?

miflower:
Perfect, that did the trick.  Thanks a lot Tomer.

PR #109, opened 2019-02-10
Fix json in response

toshetah:


azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -343,3 +345,26 @@ def test_admin_then_query(self, mock_post):
...
+        self.assertIsInstance(row[2], text_type)
+        self.assertEqual(row[2], "test bad json")
+
+        self.assertEqual(row[3], None)
+
+        self.assertIsInstance(row[4], text_type)
+        self.assertEqual(row[4], '{"rowId":2,"arr":[0,2]}')
+
+        self.assertIsInstance(row[5], dict)
+        self.assertEqual(row[5], {"rowId": 2, "arr": [0, 2]})
```
danield137: does this work? assert dict's doesn't usually work very well, I would replace it with asserting actual values inside the dict..

PR #110, opened 2019-02-10
Alter the default empty string in json according to the default response

toshetah:


azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -104,7 +104,7 @@ def test_sanity_query(self, mock_post):
             "xnumberAsText": text_type(""),
             "xtime": None,
             "xtextWithNulls": text_type(""),
-            "xdynamicWithNulls": None,
```
yifatso: why did you remove the null checks? (IIUC). can you test null dynamic values (not empty strings)?

toshetah: I am testing it at [test_dynamic](https://github.com/Azure/azure-kusto-python/blob/2aa575f1442c307ff5d7e4f021aa9d1760f1cb8b/azure-kusto-data/tests/test_kusto_client.py#L350) at test_kusto_client.py.

PR #123, opened 2019-03-25
Update helpers.py: added option to not run with increased precision for dates.

ryu577:


danield137:
Take a look at the suggested solution: https://github.com/Azure/azure-kusto-python/pull/124

I rather not introduce such a variable as it will cause data to be incompatible with Kusto Explorer.
If you strongly believe that such an escape route is useful, I will integrate it.

ryu577:
Sure, how about you let me know when you publish this change and I'll test on my query to see what the performance is like. If it continues to be an issue, the escape route might be a good option and you could print a warning when it is used saying dates will be slightly different.

danield137:
As mentioned above, linking performance fix: 

https://github.com/Azure/azure-kusto-python/tree/v0.0.28

PR #124, opened 2019-03-26
performance improvement for data frames from kusto results

danield137:


toshetah:
Let's add a check that compares the equality of 2 rows.
So we are sure we don't break it in the future.

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -48,26 +91,51 @@ def __init__(self, columns, row):
...
                             if tick:
-                                if lower_column_type == "datetime":
+                                if column_type == "datetime":
                                     self._seventh_digit[column.column_name] = tick
-                                else:
+                                elif column_type == "timespan":
                                     self._seventh_digit[column.column_name] = (
                                         tick if abs(typed_value) == typed_value else -tick
                                     )
+                                else:
```
toshetah: I am not sure that this raise is necessary, since we check the types at line 74.
But it is your call

azure-kusto-data/azure/kusto/data/_converters.py
```diff
@@ -3,19 +3,49 @@
...
 import re
 from dateutil import parser
+from dateutil.tz import UTC
 import six
 
 # Regex for TimeSpan
 _TIMESPAN_PATTERN = re.compile(r"(-?)((?P<d>[0-9]*).)?(?P<h>[0-9]{2}):(?P<m>[0-9]{2}):(?P<s>[0-9]{2}(\.[0-9]+)?$)")
 
 
+def to_high_precision_type(kusto_type, raw_value, typed_value):
```
toshetah: Maybe we want to locate this after the vanilla python converters? It will be less used.
Do you want to separate this method into two, one that takes timedelta and the other datetime?
This way every method will be consisted in its return type, and it will be easier providing names to those methods

azure-kusto-data/azure/kusto/data/_converters.py
```diff
@@ -3,19 +3,49 @@
...
+    if kusto_type == "datetime":
+        return pd.to_datetime(raw_value)
+
+    if kusto_type == "timespan":
+        if isinstance(raw_value, (six.integer_types, float)):
+            # https://docs.microsoft.com/en-us/dotnet/api/system.datetime.ticks
+            # kusto saves up to ticks, 1 tick == 100 nanoseconds
+            return pd.Timedelta(raw_value * 100, unit="ns")
+        if isinstance(raw_value, six.string_types):
+            time_parts = raw_value.split(".")
```
toshetah: Let's fix this according to what we talked offline.

azure-kusto-data/azure/kusto/data/_converters.py
```diff
@@ -3,19 +3,49 @@
...
+
+
 def to_datetime(value):
     """Converts a string to a datetime."""
     if value is None:
         return None
 
     if isinstance(value, six.integer_types):
-        return parser.parse(value)
+        parsed = parser.parse(value)
```
toshetah: Return parsed?

azure-kusto-data/azure/kusto/data/_converters.py
```diff
@@ -3,6 +3,7 @@
 from datetime import timedelta
 import re
 from dateutil import parser
+from dateutil.tz import UTC
```
toshetah: I think we can revert this file

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -9,6 +9,15 @@
 from .exceptions import KustoServiceError
 
 
+has_pandas = True
+
+try:
+    import pandas
+    from .helpers import to_pandas_datetime, to_pandas_timedelta
+except:
```
toshetah: except import error?

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,12 +1,41 @@
...
 
 
 def dataframe_from_result_table(table):
+    import pandas as pd
+    from ._models import KustoResultTable
+    from dateutil.tz import UTC
+
     """Converts Kusto tables into pandas DataFrame.
-    :param azure.kusto.data._models.KustoResultTable table: Table recieved from the response.
+    :param azure.kusto.data._models.KustoResultTable table: Table received from the response.
```
toshetah: Minor: spaces at the end

PR #125, opened 2019-03-31
Add support of streaming ingest

ArielYehezkely:
Enable streaming ingest using KustoIngestClient

### DIFF
--- a/azure-kusto-data/azure/kusto/data/request.py
+++ b/azure-kusto-data/azure/kusto/data/request.py
@@ -90,7 +90,8 @@ class KustoConnectionStringBuilder(object):
     def with_aad_user_password_authentication(cls, connection_string, user_id, password, authority_id="common"):
         """Creates a KustoConnection string builder that will authenticate with AAD user name and
         password.
-        :param str connection_string: Kusto connection string should by of the format: https://<clusterName>.kusto.windows.net
+        :param str connection_string: Kusto connection string should by of the format:
+                https://<clusterName>.kusto.windows.net
         :param str user_id: AAD user ID.
         :param str password: Corresponding password of the AAD user.
         :param str authority_id: optional param. defaults to "common"
@@ -108,7 +109,8 @@ class KustoConnectionStringBuilder(object):
     @classmethod
     def with_aad_application_key_authentication(cls, connection_string, aad_app_id, app_key, authority_id):
         """Creates a KustoConnection string builder that will authenticate with AAD application and key.
-        :param str connection_string: Kusto connection string should by of the format: https://<clusterName>.kusto.windows.net
+        :param str connection_string: Kusto connection string should by of the format:
+                https://<clusterName>.kusto.windows.net
         :param str aad_app_id: AAD application ID.
         :param str app_key: Corresponding key of the AAD application.
         :param str authority_id: Authority id (aka Tenant id) must be provided
@@ -154,7 +156,8 @@ class KustoConnectionStringBuilder(object):
     def with_aad_device_authentication(cls, connection_string, authority_id="common"):
         """Creates a KustoConnection string builder that will authenticate with AAD application and
         password.
-        :param str connection_string: Kusto connection string should by of the format: https://<clusterName>.kusto.windows.net
+        :param str connection_string: Kusto connection string should by of the format:
+                https://<clusterName>.kusto.windows.net
         :param str authority_id: optional param. defaults to "common"
         """
         kcsb = cls(connection_string)
@@ -389,7 +392,8 @@ class KustoClient(object):
 
 
 class ClientRequestProperties(object):
-    """This class is a POD used by client making requests to describe specific needs from the service executing the requests.
+    """This class is a POD used by client making requests to describe specific needs from the service executing
+    the requests.
     For more information please look at: https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties
     Not all of the documented options are implemented. You are welcome to open an issue in case you need one of them.
     """
--- a/azure-kusto-data/azure/kusto/data/request.py
+++ b/azure-kusto-data/azure/kusto/data/request.py
@@ -3,6 +3,7 @@
 import uuid
 import json
 import requests
+import io
 
 from requests.adapters import HTTPAdapter
 
@@ -262,6 +263,7 @@ class KustoClient(object):
 
         self._mgmt_endpoint = "{0}/v1/rest/mgmt".format(kusto_cluster)
         self._query_endpoint = "{0}/v2/rest/query".format(kusto_cluster)
+        self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(kusto_cluster).replace("ingest-", "")
         self._auth_provider = _AadHelper(kcsb) if kcsb.aad_federated_security else None
 
     def execute(self, database, query, properties=None):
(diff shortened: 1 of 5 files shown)


azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -324,14 +329,81 @@ def _execute(self, endpoint, database, query, default_timeout, properties=None):
...
+        :param str database: Target database.
+        :param str table: Target table.
+        :param stream: File-like object contains the data to ingest.
+        :param DataFormat stream_format: Format of the data in the stream.
+        :param str mapping_name: Pre-defined mapping of the table. Required when stream_format is json/avro.
+        Other optional params: optional request headers as documented at:
+            https://kusto.azurewebsites.net/docs/api/rest/streaming-ingest.html
+        """
+
+        assert isinstance(database, str), "Expected str instance, found {}".format(type(database))
```
danield137: we usually don't assert types, just assume it is a string (or at least behaves like one) as it doesn't really matter what type it is as long as it behaves like a string.. I would remove these..

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -324,14 +329,81 @@ def _execute(self, endpoint, database, query, default_timeout, properties=None):
...
+        assert isinstance(table, str), "Expected str instance, found {}".format(type(table))
+        assert isinstance(stream_format, str), "Expected str instance, found {}".format(type(stream_format))
+
+        request_params = {"streamFormat": stream_format}
+
+        if stream_format == "json" or stream_format == "singlejson" or stream_format == "avro":
```
danield137: simpler to read :
```suggestion
        if stream_format in ["json","singlejson","avro"]:
```

even better would be to have static var on the class itself :
`ALLOWED_FORMATS  = ["json", "avro",..]`

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -324,14 +329,81 @@ def _execute(self, endpoint, database, query, default_timeout, properties=None):
...
+            "Connection": connection or "Keep-Alive",
+            "Content-Length": str(content_length),
+            "Content-Encoding": content_encoding,
+            "Expect": expect,
+            "x-ms-client-version": "Kusto.Python.Client:" + VERSION,
+            "x-ms-client-request-id": "KPC.execute;" + str(uuid.uuid4()),
+        }
+
+        # Remove headers that are None
+        request_headers = {k: v for k, v in headers.items() if v is not None}
```
danield137: why are they `None` ? can't we make sure they aren't added? seems like a waste of time here

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -262,6 +266,7 @@ def __init__(self, kcsb):
 
         self._mgmt_endpoint = "{0}/v1/rest/mgmt".format(kusto_cluster)
         self._query_endpoint = "{0}/v2/rest/query".format(kusto_cluster)
+        self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(kusto_cluster).replace("ingest-", "")
```
danield137: should probably inline the replace (shorter string)
```suggestion
        self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(kusto_cluster.replace("ingest-", ""))
```

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -324,14 +329,81 @@ def _execute(self, endpoint, database, query, default_timeout, properties=None):
...
+
+        # Remove headers that are None
+        request_headers = {k: v for k, v in headers.items() if v is not None}
+
+        if self._auth_provider:
+            request_headers["Authorization"] = self._auth_provider.acquire_authorization_header()
+
+        request_headers["Host"] = self._streaming_ingest_endpoint.split("/")[2]
+
+        request_payload = stream
```
danield137: we don't really need these dangling variables, would just put the value int he actual call .. 

same for `endpoint` and `timeout`

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -48,63 +58,70 @@ def ingest_from_dataframe(self, df, ingestion_properties):
...
-            container_name=container_details.object_name, blob_name=blob_name, stream=descriptor.zipped_stream
-        )
-        url = blob_service.make_blob_url(container_details.object_name, blob_name, sas_token=container_details.sas)
-
-        self.ingest_from_blob(
-            BlobDescriptor(url, descriptor.size, descriptor.source_id), ingestion_properties=ingestion_properties
-        )
+        if self._use_streaming_ingest and descriptor.size < self._streaming_ingestion_size_limit:
+            if (
+                ingestion_properties.format == DataFormat.json
```
danield137: same as above,
 `ingestion_properties.format in [DataFormat.json,  DataFormat.singlejson,  DataFormat.avro]`

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -23,14 +29,18 @@ class KustoIngestClient(object):
     Tests are run using pytest.
     """
 
-    def __init__(self, kcsb):
+    def __init__(self, kcsb, use_streaming_ingest=False):
```
danield137: maybe:
```suggestion
    def __init__(self, kcsb, prefer_streaming_ingest=False):
```
?

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -48,63 +58,70 @@ def ingest_from_dataframe(self, df, ingestion_properties):
...
-
-        blob_service.create_blob_from_stream(
-            container_name=container_details.object_name, blob_name=blob_name, stream=descriptor.zipped_stream
-        )
-        url = blob_service.make_blob_url(container_details.object_name, blob_name, sas_token=container_details.sas)
-
-        self.ingest_from_blob(
-            BlobDescriptor(url, descriptor.size, descriptor.source_id), ingestion_properties=ingestion_properties
-        )
+        if self._use_streaming_ingest and descriptor.size < self._streaming_ingestion_size_limit:
```
danield137: pretty sure descriptor.size can be `None`

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -119,3 +136,58 @@ def ingest_from_blob(self, blob_descriptor, ingestion_properties):
...
         encoded = base64.b64encode(ingestion_blob_info_json.encode("utf-8")).decode("utf-8")
         queue_service.put_message(queue_name=queue_details.object_name, content=encoded)
+
+    def ingest_from_stream(self, stream, ingestion_properties):
+        """Ingest from in-memroy streams.
+        :param BytesIO or StringIO stream: in-memory stream.
+        :param azure.kusto.ingest.IngestionProperties ingestion_properties: Ingestion properties.
+        """
+
+        if isinstance(stream, io.BytesIO) or isinstance(stream, io.StringIO):
```
danield137: maybe ?
```suggestion
        if issubclass(stream, io.IOBase):
```

https://docs.python.org/3/library/io.html#io.IOBase

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -119,3 +136,58 @@ def ingest_from_blob(self, blob_descriptor, ingestion_properties):
...
+            stream.seek(0, io.SEEK_END)
```
danield137: isn't this somewhat expensive? can we assume give a stream we are always at the end of it? or at least check for that?

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -119,3 +136,58 @@ def ingest_from_blob(self, blob_descriptor, ingestion_properties):
...
+            stream_size = stream.tell()
+            stream.seek(0, io.SEEK_SET)
+        else:
+            raise ValueError("Expected BytesIO or StringIO instance, found {}".format(type(stream)))
+
+        if self._use_streaming_ingest and stream_size < self._streaming_ingestion_size_limit:
+            if (
+                ingestion_properties.format == DataFormat.json
```
danield137: again, `in` is shorter

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -12,3 +12,14 @@ class KustoDuplicateMappingError(KustoClientError):
     def __init__(self):
         message = "Ingestion properties contains ingestion mapping and ingestion mapping reference."
         super(KustoDuplicateMappingError, self).__init__(message)
+
+
+class MissingMappingReference(KustoClientError):
```
danield137: customary to suffix with Error
```suggestion
class MissingMappingReferenceError(KustoClientError):
```

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -119,3 +136,58 @@ def ingest_from_blob(self, blob_descriptor, ingestion_properties):
...
+                or ingestion_properties.format == DataFormat.singlejson
+                or ingestion_properties.format == DataFormat.avro
+            ) and ingestion_properties.mapping_reference is None:
+                raise MissingMappingReference
```
danield137: raising should be of an instance, not a class https://www.python.org/dev/peps/pep-0317/#id19
```suggestion
                raise MissingMappingReference()
```

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
```diff
@@ -119,3 +136,58 @@ def ingest_from_blob(self, blob_descriptor, ingestion_properties):
...
+            )
+
+            containers = self._resource_manager.get_containers()
+            container_details = random.choice(containers)
+            storage_client = CloudStorageAccount(
+                container_details.storage_account_name, sas_token=container_details.sas
+            )
+            blob_service = storage_client.create_block_blob_service()
+
+            # As of azure.storage.blob bug, create_blob_from_stream method fails when provided with text streams
```
danield137: can you link the bug here? so that once it's resolved we can fix this?

azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
ArielYehezkely: FileDescriptor computes the size in case of None value.

PR #127, opened 2019-04-02
Performance bug dataframes 2nd approach

danield137:


danield137:
Decided this is not the approach we want to take.
Lazy loading rows will causes equality comparator to return false

PR #136, opened 2019-04-29
Add str and repr to kusto connection string builder

toshetah:


azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -59,6 +63,27 @@ def parse(cls, key):
                 return cls.aad_federated_security
             raise KeyError(key)
 
+        def is_secret(self):
+            """States for each property if it contains secret"""
+            return self in [self.password, self.application_key, self.application_certificate]
+
+        def is_str_type(self):
+            """States whether a word is of type str or not."""
+            return self in [
```
danield137: would put this as a class var so that:
```python
return self in self.str_props
```

also, given the simple line above, would consider doping this function (condition is easy enough to read without the function)

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -59,6 +63,27 @@ def parse(cls, key):
...
+                self.application_client_id,
+                self.data_source,
+                self.password,
+                self.application_key,
+                self.authority_id,
+            ]
+
+        def is_bool_type(self):
+            """States whether a word is of type bool or not."""
+            return self in [self.aad_federated_security]
```
danield137: same as above

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -75,15 +100,31 @@ def __init__(self, connection_string):
 
         for kvp_string in connection_string.split(";"):
             key, _, value = kvp_string.partition("=")
-            self[key] = value
+            keyword = self.ValidKeywords.parse(key)
+            if keyword.is_str_type():
+                self[keyword] = value.strip()
+            if keyword.is_bool_type():
+                if value.strip() in ["True", "true"]:
```
danield137: how about we split it out to a collection (say, `truthy_values`) places in `helpers`

danield137: also, should we split this to `def ensure_type(key,value):`?

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -75,15 +100,31 @@ def __init__(self, connection_string):
...
 
     def __setitem__(self, key, value):
         try:
             keyword = key if isinstance(key, self.ValidKeywords) else self.ValidKeywords.parse(key)
         except KeyError:
             raise KeyError("%s is not supported as an item in KustoConnectionStringBuilder" % key)
 
-        self._internal_dict[keyword] = value if keyword is self.ValidKeywords.aad_federated_security else value.strip()
+        if keyword.is_str_type():
+            self._internal_dict[keyword] = value.strip()
```
danield137: this will throw if value is None, should we raise `ValueError` for  `None`? maybe as part of `ensure_type`?

azure-kusto-data/tests/test_kusto_connection_string_builder.py
```diff
@@ -8,47 +8,51 @@
 class KustoConnectionStringBuilderTests(unittest.TestCase):
     """Tests class for KustoConnectionStringBuilder."""
 
+    passwords_replacement = "****"
```
danield137: upper case const?

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -226,6 +267,23 @@ def aad_federated_security(self):
...
+        dict_copy = self._internal_dict.copy()
+        for key in dict_copy:
+            if key.is_secret():
+                dict_copy[key] = "****"
+        return self._build_connection_string(dict_copy)
+
+    def __repr__(self):
+        return self._build_connection_string(self._internal_dict)
+
+    def _build_connection_string(self, kcsb_in_dict):
```
danield137:
```suggestion
    def _build_connection_string(self, kcsb_as_dict):
```

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -226,6 +267,23 @@ def aad_federated_security(self):
...
+        result = ""
+        for word in self.ValidKeywords:
```
danield137:
```python
';'.join([ "{0}={1};".format(word.value, kcsb_as_dict[word]) for word in self.ValidKeywords if  word in kcsb_as_dict:])
```

no need for  [:-1] when using join

azure-kusto-data/azure/kusto/data/request.py
toshetah: Let's leave it here for now, and once we have more usage we can rethink where to locate it.

PR #143, opened 2019-06-03
Adding initial commit for kit-cli

danield137:
The Purpose of this project is to make ingestion simpler (one line cli, codeless ingestion)

After creating an ADX cluster via Azure, we want to explore some data.
It is a common use case while evaluating data sources or just POC-ing capabitlites we want to move fast.
That is why this project exist. It contains several common case tools to help with first steps of **schema creation** and **ingestion**.

Examples and usage are documented in the readme.

This is an initial release with basic capabilities implemented:
1. Ingestion flow for ingesting a file / folder into kusto (inferring schema, create tables, ingesting data and monitoring progress)
2. basic schema describing (from kusto, file, sql command)
3. Framework for easily** allowing different backends (storage/s3/rdbms) / description methods (cdm/hive) / Ingestion Flows to be add.


danield137:
moving to a separate repo

PR #147, opened 2019-07-04
Compress stream before sending

ArielYehezkely:
Compressing stream before uploading it to reduce network time. 
Remove size check from the client - will get error from the server.

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/_streaming_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_streaming_ingest_client.py
@@ -97,7 +97,7 @@ class KustoStreamingIngestClient(object):
             buffer = stream.read()
             with GzipFile(filename="data", fileobj=zipped_stream, mode="wb") as f_out:
                 if isinstance(buffer, string_types):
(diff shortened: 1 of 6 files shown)


azure-kusto-ingest/azure/kusto/ingest/_streaming_ingest_client.py
```diff
@@ -88,27 +86,29 @@ def ingest_from_stream(self, stream_descriptor, ingestion_properties):
...
+            zipped_stream.seek(0)
+            stream = zipped_stream
 
         self._kusto_client.execute_streaming_ingest(
             ingestion_properties.database,
             ingestion_properties.table,
             stream,
             ingestion_properties.format.name,
-            client_request_properties,
+            None,
```
danield137: I try to avoid passing `None`s, if you have position mismatch, using named params (also smart for when params aren't inferred from name, for example: `get_something(param)` - kidna obvious what param is , but `get_something(p1,p2,p3)` - is messy, much easier like `get_something(db=p1,table=p2)` ) .

ArielYehezkely: Neat =)

PR #152, opened 2019-07-17
adding support for _AadHelper pickling

zachariahcox:
In my we would like users to authenticate with aad device codes once, then make parallel kusto queries using the same client. 
A natural way to do this is via a pool map function. 
This revision allows _AadHelper to serialize via pickle.dump used by the multiprocessing libraries.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/security.py
+++ b/azure-kusto-data/azure/kusto/data/security.py
@@ -55,22 +55,25 @@ class _AadHelper(object):
 
     def __getstate__(self):
         """we cannot serialize the AuthenticationContext"""
+        state = dict(self.__dict__)
         ctx = self._adal_context
-        state = self.__dict__
-        del state['_adal_context']
-        state['adal_authority_url'] = ctx.authority.url
-        state['adal_token_cache'] = ctx.cache.serialize()
+        if ctx:
+            del state['_adal_context']
+            state['adal_authority_url'] = ctx.authority.url
+            state['adal_token_cache'] = ctx.cache.serialize()
         return state
 
     def __setstate__(self, state):
         """reconstruct the AuthenticationContext"""
-        ctx = AuthenticationContext(
-            authority=state['adal_authority_url'],
-            cache = TokenCache(state['adal_token_cache']))
-        del state['adal_authority_url']
-        del state['adal_token_cache']
+        authority_url = state.get('adal_authority_url')
+        if authority_url:
+            del state['adal_authority_url']
+        token_cache_string = state.get('adal_token_cache')
+        if token_cache_string:
+            del state['adal_token_cache']
+        if authority_url and token_cache_string:
+            state['_adal_context'] = AuthenticationContext(authority_url, cache=TokenCache(token_cache_string))
         self.__dict__ = state
-        self._adal_context = ctx
 
     def acquire_authorization_header(self):
         """Acquire tokens from AAD."""
--- a/azure-kusto-data/azure/kusto/data/security.py
+++ b/azure-kusto-data/azure/kusto/data/security.py
@@ -8,13 +8,14 @@ import dateutil.parser
 
 from adal import AuthenticationContext, AdalError
 from adal.constants import TokenResponseFields, OAuth2DeviceCodeResponseParameters
+from adal.token_cache import TokenCache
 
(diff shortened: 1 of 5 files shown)


danield137:
I'm not sure I understand what is the scenario you with to support? Please describe it.
If you could share a code snippet it would be great!

In general, I dislike pickling, because pickles are very fragile, and, I have no intention of them being passed around outside the scope of the client.

If the scenario is multi-threaded environment, then we have not tested our sdk, and other stuff might break, so it will require more work.

zachariahcox:
I'll add some comments inline, but I added a unit test that covers our main use case -- using Pool.map and surviving the pickle call they use to get the client into other threads.

azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -389,3 +390,21 @@ def test_empty_result(self, mock_post):
...
+
+    @pytest.mark.skipif(sys.version_info < (3, 6), reason="requires python3.6 or higher")
+    def test_running_in_parallel(self):
+        """authenticated clients should be able to do things in parallel"""
+        from multiprocessing import Pool
+        cluster = "https://somecluster.kusto.windows.net"
+        client = KustoClient(cluster)
+        inputs = [client for i in range(5)]
+        with Pool(2) as p:
+            responses = p.map(getendpoint, inputs)
```
zachariahcox: @danield137, this basically covers it -- the api we're trying to call in our app is KustoClient.execute (which appears to turn around and use requests). 
This fix just lets the client object become accessible from the pool which implicitly uses pickle for transport.

azure-kusto-data/tests/test_security.py
```diff
@@ -22,3 +22,14 @@ def test_unauthorized_exception():
         assert error.authority == "https://login.microsoftonline.com/authorityName"
         assert error.kusto_cluster == cluster
         assert error.kwargs["username"] == username
+
+def test_pickle():
```
zachariahcox: I can also rename this test if you like -- the real intent here is extremely short-term persistence via multiprocessing.Pool.map's pickle for moving objects to other threads.

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -52,6 +53,28 @@ def __init__(self, kcsb):
...
+    def __setstate__(self, state):
+        """reconstruct the AuthenticationContext"""
+        authority_url = state.get('adal_authority_url')
+        if authority_url:
+            del state['adal_authority_url']
+        token_cache_string = state.get('adal_token_cache')
+        if token_cache_string:
+            del state['adal_token_cache']
+        if authority_url and token_cache_string:
+            state['_adal_context'] = AuthenticationContext(authority_url, cache=TokenCache(token_cache_string))
```
zachariahcox: This is the critical line -- when coming back from the pickle, we reconstruct the thread-safe adal.AuthenticationContext using the original process' token cache

In our app, we want users to be able to define functions that they can run with multiprocessing.Pool.map that accept KustoClient objects for making queries. 
The simplest case would look like this: 

```python
def f(data, client):
    query = do_a_bunch_of_work_to_create_my_query(data)
    result = client.execute(query)
    return do_a_little_more_work(result)

client = KustoClient(...)  # use aad device auth
client.execute('db', 'print 1 | take 0') # run a query to force authentication
with Pool() as p:
    results = p.starmap(f, [(data, client) for data in lots_of_datas])
```

zachariahcox:
> `AuthenticationContext` is not pickleable.

It is not actually important that this class be serializable -- it has rlocks and things. 
It _is_ important that the KustoClient be serializable, which means at least being constructible with a predefined token cache. 

We have actually run into another use case for something like this -- we'd like to streamline the kusto authentication process for when users need to auth with aad device login for multiple clusters. 

If we can acquire the aad tokens using the adal libraries ourselves, we would like to be able to construct kusto clients with these tokens caches directly.

yogilad:
Hi @zachariahcox , 

Apologies for the very late response to this.

We feel this code does not belong in the Python SDK itself. 
It could be a standalone solution that uses Kusto Python SDK, but it would require significant changes to become a reliable logging framework (persistent caching, etc.).

We do support LogStash and plan to support Open Telemetry in the future. This would provide a reliable logging framework without the need to address the subject in the Kusto section of the code.

PR #160, opened 2019-08-25
Deals with non nullable values

boazsha:
* Deals with None values
* Generalized calls for Pandas conversion functions

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_models.py
+++ b/azure-kusto-data/azure/kusto/data/_models.py
@@ -30,11 +30,9 @@ class WellKnownDataSet(Enum):
 class KustoResultRow(object):
     """Iterator over a Kusto result row."""
 
-    convertion_funcs = {
-        "datetime": _converters.to_datetime,
-        "timespan": _converters.to_timedelta,
-        "decimal": Decimal
-    }
+    convertion_funcs = {"datetime": _converters.to_datetime, "timespan": _converters.to_timedelta, "decimal": Decimal}
+
+    pandas_funcs = {"datetime": to_pandas_datetime, "timespan": to_pandas_timedelta}
 
     def __init__(self, columns, row):
         self._value_by_name = {}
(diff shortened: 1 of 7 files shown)


azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -48,31 +50,29 @@ def __init__(self, columns, row):
...
+                )
+
+                # This is a special case where plain python will lose precision, so we keep the precise value hidden.
+                # When transforming to pandas, we can use the hidden value to convert to precise pandas/numpy types
                 if HAS_PANDAS:
-                    self._hidden_values.append(value)
+                    self._hidden_values.append(
+                        KustoResultRow.pandas_funcs[column_type](value, typed_value)
+                        if column_type in KustoResultRow.pandas_funcs
+                        else value
```
danield137: shouldn't this `else` be `typed_value`?

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -48,31 +50,29 @@ def __init__(self, columns, row):
...
-                    # this is a special case where plain python will lose precision, so we keep the precise value hidden
-                    # when transforming to pandas, we can use the hidden value to convert to precise pandas/numpy types
-                    if HAS_PANDAS:
-                        if column_type == "datetime":
-                            self._hidden_values.append(to_pandas_datetime(value))
-                        if column_type == "timespan":
-                            self._hidden_values.append(to_pandas_timedelta(value, typed_value))
-                        if column_type == "decimal":
-                            self._hidden_values.append(value)
+            if value is None:
```
danield137: how about you get rid of the if else...

```python
# If you are here to read this, you probably hit some datetime/timedelta inconsistencies.
# Azure-Data-Explorer(Kusto) supports 7 decimal digits, while the corresponding python types supports only 6.
# One example why one might want this precision, is when working with pandas.
# In that case, use azure.kusto.data.helpers.dataframe_from_result_table which takes into account the original value.
typed_value = (
    KustoResultRow.convertion_funcs[column_type](value)
    if column_type in KustoResultRow.convertion_funcs and value is not None
    else value
)

    # This is a special case where plain python will lose precision, so we keep the precise value hidden.
# When transforming to pandas, we can use the hidden value to convert to precise pandas/numpy types
if HAS_PANDAS:
    self._hidden_values.append(
        KustoResultRow.pandas_funcs[column_type](value, typed_value)
        if column_type in KustoResultRow.pandas_funcs and value is not None
        else typed_value 
    )
```

azure-kusto-data/tests/test_converter.py
```diff
@@ -10,8 +10,7 @@ class ConverterTests(unittest.TestCase):
 
     def test_to_timestamp(self):
         """Happy path to test converter from TimeSpan to timedelta."""
-        # Test None
-        self.assertIsNone(to_timedelta(None))
+        # @Daniel
```
danield137: remove comments 👍 

you can tag me like so @boazsha

azure-kusto-data/azure/kusto/data/_models.py
boazsha: [Here](https://github.com/Azure/azure-kusto-python/blob/e626d40cdb079d717d05133b9bde301bb071e73b/azure-kusto-data/azure/kusto/data/_models.py#L71) and [here](https://github.com/Azure/azure-kusto-python/blob/6bf07c3c82c00ccb6ec2bf52711e0723dfbf71de/azure-kusto-data/azure/kusto/data/_models.py#L73) value is appended (which is a string representing decimal). Should it append typed_value?

azure-kusto-data/azure/kusto/data/_models.py
boazsha: This was my initial implementation (see [commit](https://github.com/Azure/azure-kusto-python/blob/564b972bd36e453fd87c43092f1263b0d7af28f9/azure-kusto-data/azure/kusto/data/_models.py#L61)), but I think this way is cleaner (and more efficient with regards to function calls). Let me know what you think.

danield137: How is it more efficient? you can always reverse 
`if value is not None and column_type in KustoResultRow.pandas_funcs`
which will do exactly the same. 

My problem with the `if value is Non` is that it doesn't convey the intention of the code as well as the code without it..

But, it's not that big of a deal to me.. if you like it the way it was, ok.

PR #166, opened 2019-09-17
Updating file creation logic for ingestion.

sridharan99:
If Multiple requests are sent at same second, the file is getting deleted during the first request but the second request is failing with FileNotFoundError as we are trying to delete files for each request(Line No. 69). Using guid should solve the problem.

Error log. 

2019-09-17T13:52:56.387267791Z FileNotFoundError: [Errno 2] No such file or directory: '/tmp/df_1568728375_10.csv.gz'
2019-09-17T13:52:55.528903798Z FileNotFoundError: [Errno 2] No such file or directory: '/tmp/df_1568728375_10.csv.gz'

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_version.py
+++ b/azure-kusto-data/azure/kusto/data/_version.py
@@ -1,2 +1,2 @@
 """Version file."""
-VERSION = "0.0.33"
+VERSION = "0.0.34"
--- a/azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py
@@ -4,6 +4,7 @@ import base64
 import random
 import uuid
 import os
+import time
 import tempfile
 
 from azure.storage.common import CloudStorageAccount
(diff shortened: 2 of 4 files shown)


danield137:
How does it happen? Do you try to ingest different data frames at the same time on the same process? How? 
In general, fix looks OK, but I would keep the timestamp and the guid, timestamp is useful for debugging..

sridharan99:
> How does it happen? Do you try to ingest different data frames at the same time on the same process? How?
> In general, fix looks OK, but I would keep the timestamp and the guid, timestamp is useful for debugging..

Yes. Multiple requests were sent at same time. 

Error log
2019-09-17T13:52:56.387267791Z FileNotFoundError: [Errno 2] No such file or directory: '/tmp/df_**1568728375**_10.csv.gz'
2019-09-17T13:52:55.528903798Z FileNotFoundError: [Errno 2] No such file or directory: '/tmp/df_**1568728375**_10.csv.gz'

It will be useful to have timestamp as you are suggesting. Pushed an iteration.

danield137:
How about using id(data frame)instead of guid? I want to make sure that writing the exact same data  will still cause an error

sridharan99:
The data ingestion should be controlled by the client application which is
using this library. The method should be generic enough to handle the data
ingestion.

danield137:
@sridharan99 I'm not sure I understand how you answer relates to my statement. I don't think ingesting the same dataframe (we can know that it is the same, by using `id`) from the same process, at the same time is something a user would like to do, and I assume it's a mistake. That is why I prefer to use `id(dataframe)` over guid. 

If you wish to let the client control this, we should expose a way for him to pass an identifier for each dataframe ingestion. We already do so for other ingestion methods by allowing to set a `source_id` on `ingestion_properties`, and respecting it as a unique id.

So, in essence, what I propose is to do something like [this](https://github.com/Azure/azure-kusto-python/blob/8d10ae0c0ae38e29cc47a55f662b2e6779601cbf/azure-kusto-ingest/azure/kusto/ingest/_ingest_client.py#L88), providing a way to explicitly set the uid by the end user:

This will require more effort to do, as it will require you to either add a new descriptor, or, the smarter approach would be to have a base descriptor to only hold the source_id property.
After which, you can add that descriptor to the method signature...

PR #168, opened 2019-09-19
Added support for ingestionMappingType and ingestionMappingReference

boazsha:
Added support for IngestionProperties: ingestionMappingType and ingestionMappingReference

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
@@ -43,18 +43,18 @@ class _IngestionBlobInfo:
             additional_properties["ingestIfNotExists"] = _convert_list_to_json(
                 ingestion_properties.ingest_if_not_exists
             )
-        if ingestion_properties.mapping:
-            json_string = _convert_dict_to_json(ingestion_properties.mapping)
-            additional_properties[ingestion_properties.get_mapping_format() + "Mapping"] = json_string
+        if ingestion_properties.ingestion_mapping:
+            json_string = _convert_dict_to_json(ingestion_properties.ingestion_mapping)
+            additional_properties["ingestionMapping"] = json_string
 
         if ingestion_properties.ingestion_mapping_reference:
             additional_properties["ingestionMappingReference"] = ingestion_properties.ingestion_mapping_reference
         if ingestion_properties.ingestion_mapping_type:
-            additional_properties["ingestionMappingType"] = ingestion_properties.ingestion_mapping_type.name
+            additional_properties["ingestionMappingType"] = ingestion_properties.ingestion_mapping_type.value
         if ingestion_properties.validation_policy:
             additional_properties["ValidationPolicy"] = _convert_dict_to_json(ingestion_properties.validation_policy)
         if ingestion_properties.format:
-            additional_properties["format"] = ingestion_properties.format.name
+            additional_properties["format"] = ingestion_properties.format.value
 
         if additional_properties:
             self.properties["AdditionalProperties"] = additional_properties
--- a/azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
@@ -47,7 +47,7 @@ class _IngestionBlobInfo:
             json_string = _convert_dict_to_json(ingestion_properties.mapping)
             additional_properties[ingestion_properties.get_mapping_format() + "Mapping"] = json_string
 
-        # mapping_reference will be
+        # mapping_reference will be deprecated in the following versions
         if ingestion_properties.mapping_reference:
             additional_properties["ingestionMappingReference"] = ingestion_properties.mapping_reference
         elif ingestion_properties.ingestionMappingReference:
--- a/azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
@@ -49,6 +49,10 @@ class _IngestionBlobInfo:
         if ingestion_properties.mapping_reference:
             key = ingestion_properties.get_mapping_format() + "MappingReference"
             additional_properties[key] = ingestion_properties.mapping_reference
+        if ingestion_properties.ingestionMappingType:
+            additional_properties["ingestionMappingType"] = ingestion_properties.ingestionMappingType.name
+        if ingestion_properties.ingestionMappingReference:
+            additional_properties["ingestionMappingReference"] = ingestion_properties.ingestionMappingReference
         if ingestion_properties.validation_policy:
             additional_properties["ValidationPolicy"] = _convert_dict_to_json(ingestion_properties.validation_policy)
         if ingestion_properties.format:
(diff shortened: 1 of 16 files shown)


azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, blob_descriptor, ingestion_properties, auth_context=None):
         if ingestion_properties.mapping_reference:
```
ArielYehezkely: IMHO we should choose either "mapping_reference" or "ingestionMappingReference" as both have the same use. If we are changing the name we should declare a breaking change.

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
```diff
@@ -21,6 +21,20 @@ class DataFormat(Enum):
     multijson = "multijson"
 
 
+class IngestionMappingType(Enum):
```
ArielYehezkely: The types are: Csv, Json, Avro, Parquet

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
```diff
@@ -135,5 +155,7 @@ def get_mapping_format(self):
             return DataFormat.json.name
         elif self.format == DataFormat.avro:
             return DataFormat.avro.name
+        elif self.format == DataFormat.parquet:
```
ArielYehezkely: You can remove this method and use the new property "ingestionMappingReference"

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
```diff
@@ -21,6 +21,20 @@ class DataFormat(Enum):
     multijson = "multijson"
 
 
+class IngestionMappingType(Enum):
+    Csv = "Csv"
```
danield137: we should fix this across, but python seems to favour upper case naming convention for enums 
https://docs.python.org/3/library/enum.html#creating-an-enum

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -10,7 +10,10 @@ class KustoDuplicateMappingError(KustoClientError):
     """
 
     def __init__(self):
-        message = "Ingestion properties contains ingestion mapping and ingestion mapping reference."
+        message = (
```
danield137: you meant to use 
```
"""
This is multi line string in python
"""
```

azure-kusto-ingest/pyproject.toml
```diff
@@ -1,2 +0,0 @@
-[tool.black]
```
danield137: I am not sure why removed this.

azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, blob_descriptor, ingestion_properties, auth_context=None):
...
             key = ingestion_properties.get_mapping_format() + "MappingReference"
             additional_properties[key] = ingestion_properties.mapping_reference
+        if ingestion_properties.ingestionMappingType:
```
danield137: python naming convention is snake_case

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
boazsha: Using this with several lines will result in a string that includes newlines, which - if I understand correctly - is not desirable in this case (printed out). What do you think?

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
boazsha: This method is also used with the property `mapping`. I'll see if I can remove it as well

azure-kusto-ingest/azure/kusto/ingest/_ingestion_blob_info.py
boazsha: I do agree with you that the propery to be used when constructing `IngestionBlobInfo` is `ingestionMappingReference`. Thanks for that.
But, with regards to removing `mapping_reference` completely, won't a better solution be to issue a deprecation warning and to remove it a couple of versions down the road? (see latest commit for both fixes)

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
```diff
@@ -112,13 +122,25 @@ def __init__(
         validationPolicy=None,
         additionalProperties=None,
     ):
-        if mapping is not None and mappingReference is not None:
+        # mapping_reference will be deprecated in the following versions
+        if mappingReference is not None:
+            warnings.warn(
+                "mappingReference will be deprecated in the following versions."
```
danield137: multi line strings with `"""`

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
```diff
@@ -112,13 +122,25 @@ def __init__(
...
+                " Please use ingestionMappingReference instead",
+                PendingDeprecationWarning,
+            )
+        if (mapping is not None and (mappingReference is not None or ingestionMappingReference is not None)) or (
```
danield137: this is hard to read (and i think is wring as well, considering you have the same condition twice).
break it down into multiple, logically distinguishable ifs (if needed, use vars to make it readable , like `mapping_exists = mapping is not None or mappingReference is not None`)

azure-kusto-ingest/azure/kusto/ingest/_streaming_ingest_client.py
```diff
@@ -19,7 +19,7 @@ class KustoStreamingIngestClient(object):
     Tests are run using pytest.
     """
 
-    _mapping_required_formats = [DataFormat.json, DataFormat.singlejson, DataFormat.avro]
+    _mapping_required_formats = [DataFormat.JSON, DataFormat.SINGLEJSON, DataFormat.AVRO]
```
danield137: although negligible, as long as you interned to lookup values (and not iterate), sets are more efficient. `{DataFormat.JSON, DataFormat.SINGLEJSON, DataFormat.MULTIJSON}`

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
danield137: I think I expect to read code string, just as they will appear in real life. if you intend it to be a single line comment, just write it in a single line.

You will have to deal with linters, which we currently set to 120 chars. 
Do you think 120 chars applies to comments as well as code (hint: I do) ? If so, it should be multiline, other wise, use linting exclusions...

https://black.readthedocs.io/en/stable/the_black_code_style.html
> Black reformats entire files in place. It is not configurable. It doesn’t take previous formatting into account. It doesn’t reformat blocks that start with # fmt: off and end with # fmt: on

azure-kusto-ingest/azure/kusto/ingest/_ingestion_properties.py
```diff
@@ -1,24 +1,32 @@
...
+    TXT = "txt"
+    JSON = "json"
+    SINGLEJSON = "singlejson"
+    AVRO = "avro"
+    PARQUET = "parquet"
+    MULTIJSON = "multijson"
+
+
+class IngestionMappingType(Enum):
+    CSV = "Csv"
```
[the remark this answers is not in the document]
boazsha: Yes

PR #183, opened 2019-11-05
Fix spelling - "available"

scovetta:


README.md
```diff
@@ -12,10 +12,10 @@
 
 https://github.com/Azure/azure-kusto-ingestion-tools/tree/master/kit
 
-[*azure-kusto-ingestion-tools*]("https://github.com/Azure/azure-kusto-ingestion-tools/tree/master/kit")  a simple toolkit to help with ingestions, avialble here
-<br>
```
ArielYehezkely: I would have removed all the line breaks or none (I agree this one should not be in a separate line)

PR #190, opened 2019-12-09
Remove unnecessary lines from dataframe helper function

victorlin:
- `dateutil.tz.UTC` is never used - looks like an artifact from a removed check for datetime.
- Seems like bool data type is already handled correctly in `KustoResultTable`.


victorlin:
That's odd, the test still passes on my local machine even after removing the bool check. Not sure why it fails on the CI build.

In any case, the import line can still be removed.

danield137:
Does your machine have pandas installed on that specific virtual env? Are you sure the tests related to pandas run on your machine?

Regarding the import, looks like you are correct, if you want to change the PR to only include it, I will approve

victorlin:
Nevermind, I ran the tests again and they failed. Will update the PR.

danield137:
Merged. Thanks for your contribution!

Issue #153, opened 2019-07-19
Suggestion: add support for async (python > 3.6)


danield137:

vladikbr:
Change committed

PR #202, opened 2020-02-08
[FEAT] add aio support

mordajon:
*Async queries is done natively using aiohttp behind the scenes*
*Async authentication is done using ThreadExecutor due to library dependencies*

1. Added aio API and transform all derivation to be none blocking
2. Previous sync API left the same way while extracting common logic into reusable functions
3. Added dependency in aiohttp, asgiref, aiofile when installing optional async
   - aiohttp (for native async http calls)
   - asgiref (msrestazure, adal, azure-cli - are blocking libraries, so we need to wrap all of them in executor)
   - aiofile (azure cli authentication reads from file, doing it async now)
4. Added static typings
5. Added async installation to travis
6. Async support was added according to [azure guidelines](https://azure.github.io/azure-sdk/python_design.html#async-support)
7. Modified ReadMe to show examples of usage
8. Added helper decorators to avoid documentation duplication
9. Modified .travis.yml matrix


mordajon:
Usage example:
image

Output Example:
image

danield137:
https://github.com/Azure/azure-kusto-python/issues/153

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -506,6 +512,7 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
         kusto_cluster = kcsb.data_source
 
         # Create a session object for connection pooling
+        # self._session is only used using sync access
```
danield137: why? can't use session in async flow?

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -516,7 +523,20 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
         self._auth_provider = _AadHelper(kcsb) if kcsb.aad_federated_security else None
         self._request_headers = {"Accept": "application/json", "Accept-Encoding": "gzip,deflate", "x-ms-client-version": "Kusto.Python.Client:" + VERSION}
 
+    @staticmethod
+    def _build_runtime_error(func_name: str) -> RuntimeError:
```
danield137: I dislike build functions. these are ctors, and should be implemented as such. Also, since this is an expected error, we should probably raise it with a `Kusto` prefix like `KustoAsyncRuntimeError`

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -516,7 +523,20 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
...
+        return RuntimeError(f"Kusto driver has detected a running event loop, consider using {func_name}_async instead of {func_name}")
+
     def execute(self, database: str, query: str, properties: ClientRequestProperties = None) -> KustoResponseDataSet:
+        """
+        Sync version of self.execute_async
+        """
+        try:
+            return async_to_sync(self.execute_async)(database, query, properties)
+        except RuntimeError as e:
+            raise self._build_runtime_error("execute") from e
```
danield137: this means that running sync in a non-async env would break?

azure-kusto-data/azure/kusto/data/_version.py
```diff
@@ -1,2 +1,2 @@
 """Version file."""
-VERSION = "0.0.43"
+VERSION = "0.0.44"
```
danield137: this is only done when releasing a version.. no need

azure-kusto-data/azure/kusto/data/asyncio/utils.py
```diff
@@ -0,0 +1,9 @@
+import aiohttp
+
+
+async def run_request(endpoint: str, **kwargs) -> dict:
```
danield137: I dislike utils classes. I prefer you put this function close to it's usage (kusto doesn't provide utils)

azure-kusto-data/setup.py
```diff
@@ -41,6 +41,6 @@
     namespace_packages=["azure"],
     keywords="kusto wrapper client library",
     packages=find_packages(exclude=["azure", "tests"]),
-    install_requires=["adal>=1.0.0", "python-dateutil>=2.8.0", "requests>=2.13.0", "msrestazure>=0.4.14"],
+    install_requires=["adal>=1.0.0", "python-dateutil>=2.8.0", "requests>=2.13.0", "msrestazure>=0.4.14", "aiohttp==3.6.2", "asgiref==3.2.3", "aiofile==1.5.2"],
```
danield137: these should be optionals. (like the pandas thing). people using sync flow don't need them.

azure-kusto-data/azure/kusto/data/request.py
mordajon: Nope... it means that when trying to run sync flow from async env it would break...

mordajon:
Tests will be added upon API approval by @danield137

azure-kusto-data/azure/kusto/data/request.py
mordajon: Async operations are wrapped by aiohttp.ClientSession which is way way faster:
image
https://julien.danjou.info/python-and-fast-http-clients/

azure-kusto-data/azure/kusto/data/request.py
mordajon: 👍

azure-kusto-data/azure/kusto/data/_version.py
mordajon: 👍

azure-kusto-data/setup.py
mordajon: Ammm... I am not sure I am following you with this? why are they optional?

azure-kusto-data/azure/kusto/data/asyncio/utils.py
mordajon: 👍

azure-kusto-data/setup.py
danield137: In theory, installing these packages takes time and space.
TBH, I'm still not sure I want async flow to be the default right now.

I want people to **opt-in** into this flow.
This means that people that want to use the regular flow, still can without extra cost.

.travis.yml
```diff
@@ -17,7 +17,7 @@ install:
   - pip install -U pip
   - pip install -r dev_requirements.txt
   - pip install codecov coverage
-  - pip install ./azure-kusto-data ./azure-kusto-ingest
+  - pip install ./azure-kusto-data[async] ./azure-kusto-ingest
```
danield137: Please revert.
Testing the sync flow without dependencies is more important.
That being said, we need to come up with a way to test both scenarios independently.

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -530,6 +546,18 @@ def execute(self, database: str, query: str, properties: ClientRequestProperties
             return self.execute_mgmt(database, query, properties)
         return self.execute_query(database, query, properties)
 
+    async def execute_async(self, database: str, query: str, properties: ClientRequestProperties = None) -> KustoResponseDataSet:
```
danield137: how does python handle the async keyword if no asynio is installed?

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -116,32 +150,61 @@ def __init__(self, kcsb):
         aad_authority_uri = os.environ.get("AadAuthorityUri", CLOUD_LOGIN_URL)
         self.authority_uri = aad_authority_uri + authority if aad_authority_uri.endswith("/") else aad_authority_uri + "/" + authority
 
-    def acquire_authorization_header(self):
+    def _build_kusto_authentication_error(self, error: KustoAuthenticationError) -> KustoAuthenticationError:
```
danield137: should be a `KustoAuthenticationError` ctor / factory method

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -198,15 +253,81 @@ def _acquire_authorization_header(self) -> str:
...
+        if self.authentication_method == AuthenticationMethod.az_cli_profile:
+            stored_token = await _get_azure_cli_auth_token_async()
+            self._set_from_stored_token(stored_token)
+            refresh_token = stored_token[TokenResponseFields.REFRESH_TOKEN]
+
+        if self.auth_context is None:
+            self.auth_context = AuthenticationContext(self.authority_uri)
+
+        if refresh_token is not None:
+            token = await sync_to_async(self.auth_context.acquire_token_with_refresh_token)(refresh_token, self.client_id, self.kusto_uri)
```
danield137: I'm not sure this is safe. Adal's AuthContext was not built with multi-threading in mind.

azure-kusto-data/tests/async_unittest.py
```diff
@@ -0,0 +1,11 @@
+import unittest
```
danield137: please use pytest (unitest code is legacy code)

azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -341,43 +327,186 @@ def test_admin_then_query(self, mock_post):
...
         assert row[3] is None
-
         assert isinstance(row[4], str)
         assert row[4] == '{"rowId":2,"arr":[0,2]}'
-
         assert isinstance(row[5], dict)
         assert row[5] == {"rowId": 2, "arr": [0, 2]}
 
+
+class KustoClientTestsSync(unittest.TestCase, KustoClientMixin):
```
danield137: Please revert this. I'm not a fan of splitting assertion code from tests.
There are other ways of reducing code.
(also, this is unrelated to the PR)

azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -341,43 +327,186 @@ def test_admin_then_query(self, mock_post):
...
+        query = """print dynamic(123), dynamic("123"), dynamic("test bad json"),"""
+        """ dynamic(null), dynamic('{"rowId":2,"arr":[0,2]}'), dynamic({"rowId":2,"arr":[0,2]})"""
+        row = client.execute_query("PythonTest", query).primary_results[0].rows[0]
+        self._assert_dynamic_response(row)
+
     @patch("requests.Session.post", side_effect=mocked_requests_post)
     def test_empty_result(self, mock_post):
         """Tests dynamic responses."""
-        client = KustoClient("https://somecluster.kusto.windows.net")
+        client = self._create_kusto_client()
```
danield137: revert. I believe in explicit creation over implicit, this code hides the actual value for no real reason.

azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -341,43 +327,186 @@ def test_admin_then_query(self, mock_post):
...
+        self._mock_query(mocked)
+        client = self._create_kusto_client()
+        response = self.loop.run_until_complete(client.execute_query_async("PythonTest", "Deft"))
+        self._assert_sanity_query_response(response)
+
+    @aioresponses()
+    def test_sanity_control_command_async(self, mocked):
+        """Async version of self.test_sanity_control_command"""
+        self._mock_mgmt(mocked)
+        client = self._create_kusto_client()
```
danield137: pytest (and me) are against test classes. you should refrain from using `self` in tests

dev_requirements.txt
```diff
@@ -2,4 +2,6 @@ pytest>=3.2.0
 mock>=2.0.0
 responses>=0.9.0
 pandas>=0.24.0
-black;python_version == '3.6'
\ No newline at end of file
+black;python_version == '3.6'
+aioresponses==0.6.2
```
danield137: like the above, and previous discussion, I do not intend for asyn flow to be the main flow. as such, we cannot declare it as an equal rights citizen everwhere. it should always be Opt-In.

danield137:
After the last changes, I'm starting to think a better way would be to just have a separate class entirely.

`KustoClient` vs `KustoAsyncClient`

This will create the cleanest separation possible.

Also, one thing I did not consider is that for async flow to work well, everything needs to be async.
Since that is not the case (adal) I am not sure what are the implication of faking it.

mordajon:
> After the last changes, I'm starting to think a better way would be to just have a separate class entirely.
> 
> `KustoClient` vs `KustoAsyncClient`
> 
> This will create the cleanest separation possible.
> 
> Also, one thing I did not consider is that for async flow to work well, everything needs to be async.
> Since that is not the case (adal) I am not sure what are the implication of faking it.

A lot of libraries out there wrapping sync actions with thread executor. 
We only wrap the authentication which happens once per... 

Also I am working with ADAL team as well to create an async flow there as well.
We can use it in the next version 😊

azure-kusto-data/azure/kusto/data/request.py
mordajon: Asyncio comes integrated with python 3.5+

azure-kusto-data/tests/async_unittest.py
mordajon: I agree, pytest is way better... But I wanted to keep the same writing style

azure-kusto-data/tests/test_kusto_client.py
mordajon: It is related because async flow uses the same asserts

dev_requirements.txt
mordajon: What is your suggestion in regards of dev_requirements?

mordajon: I believe in tests we should check all of the flows, don't you think?

.travis.yml
mordajon: Amm well I think that tests should cover everything... to prevent code changes that will break async or panda for that matter

azure-kusto-data/azure/kusto/data/security.py
mordajon: From asgi docs:
"If the outermost program is async (i.e. SyncToAsync is outermost), then
    this will be a dedicated single sub-thread that all sync code runs in,
    one after the other"

azure-kusto-data/tests/test_kusto_client.py
mordajon: I agree that pytest is the better approach...

But the way I see it, we either change test_kusto_client to be written in pytest completely or keep it the way it is today.

My thought was that you won't want to see a lot of changes in the tests (to make sure everything runs exactly like before)

dev_requirements.txt
mordajon: do mind... that pandas is in that file as well

danield137: there is a large difference between the two, but I agree that pandas should also be out of these tests.

mordajon:
Creating another commit to follow Azure guidelines:
https://azure.github.io/azure-sdk/python_design.html#async-support

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -16,6 +18,18 @@
...
     HAS_PANDAS = False
 
+aio_installed = True
+try:
+    from aiohttp import ClientResponse, ClientSession
+except ImportError:
+    aio_installed = False
+
+
+def _get_dynamic_response_type():
```
danield137: where do you use this?

plus, not sure why you need this code here, everything aio is under subfolder

azure-kusto-data/azure/kusto/data/aio/request.py
```diff
@@ -0,0 +1,75 @@
...
+
+try:
+    from aiohttp import ClientResponse, ClientSession
+except ImportError:
+    raise KustoAioSyntaxError()
+
+ResponseTuple = namedtuple("ResponseTuple", ["response", "response_json"])
+
+
+@documented_by(KustoClientSync)
```
danield137: please don't do this.
it blocks the ability to render this file to a readthedocs page. comments should be static

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -1,631 +1,657 @@
...
-            return KustoResponseDataSetV1(response.json())
-
-        if payload:
-            raise KustoServiceError(
-                "An error occurred while trying to ingest: Status: {0.status_code}, Reason: {0.reason}, Text: {0.text}".format(response), response
-            )
-
-        raise KustoServiceError([response.json()], response)
+"""A module to make a Kusto request."""
+import io
```
danield137: did you change line endings or something? the entire file shouldn't be changed

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -1,631 +1,657 @@
...
+                request_headers["x-ms-app"] = properties.application
+            if properties.user is not None:
+                request_headers["x-ms-user"] = properties.user
+        if properties:
+            timeout = properties.get_option(ClientRequestProperties.request_timeout_option_name, timeout)
+
+        return RequestParams(json_payload, request_headers, timeout)
+
+    @staticmethod
+    def _build_kusto_service_error(payload: io.IOBase, response: _get_dynamic_response_type(), response_json: List[dict]) -> KustoServiceError:
```
danield137: already commented about this: this should be a ctor for the kustoserviceerror

danield137: this `_get_dynamic_response_type` beats the purpose of annotations. annotations are here to help me read the code and make sure I don't make typing mistakes. if they are resolved in runtime it does not help at all.
please revert this.

if you want a dynamic type, just use `Any` and duck type it

azure-kusto-data/azure/kusto/data/request.py
```diff
@@ -1,631 +1,657 @@
...
+    def get_option(self, name, default_value):
+        """Gets an option's value."""
+        return self._options.get(name, default_value)
+
+    def to_json(self):
+        """Safe serialization to a JSON string."""
+        return json.dumps({"Options": self._options, "Parameters": self._parameters}, default=str)
+
+
+RequestParams = namedtuple("RequestParams", ["json_payload", "request_headers", "timeout"])
```
danield137: underscore (mark as private)

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -116,57 +79,100 @@ def __init__(self, kcsb):
...
+            self.client_id = stored_token[TokenResponseFields._CLIENT_ID]
+            self.username = stored_token[TokenResponseFields.USER_ID]
+            self.authority_uri = stored_token[TokenResponseFields._AUTHORITY]
+
+    @staticmethod
+    def _get_expiration_data_from_token(token):
+        return dateutil.parser.parse(token[TokenResponseFields.EXPIRES_ON])
+
+    @staticmethod
+    def _authentication_method_missing_exception() -> KustoClientError:
```
danield137: don't wrap error creation. all over the place.

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -116,57 +79,100 @@ def __init__(self, kcsb):
...
+        return KustoClientError("Please choose authentication method from azure.kusto.data.security.AuthenticationMethod")
+
+    @staticmethod
+    def _is_expired(expiration_date):
```
danield137: don't hide the logic. nobody knows what expired means. this method should be removed

azure-kusto-data/tests/aio/test_kusto_client.py
```diff
@@ -0,0 +1,127 @@
...
+    import asgiref
+
+    aio_installed = True
+except:
+    pass
+
+
+@pytest.mark.skipif(not aio_installed, reason="requires aio")
+@aio_documented_by(KustoClientTestsSync)
+class KustoClientTestsTests(TestCase, KustoClientTestsMixin):
```
danield137: naming? why the double tests?

azure-kusto-data/azure/kusto/data/security.py
mordajon: I think this is beyond of the scope of this task, and this PR is very big already... let's do it later

azure-kusto-data/azure/kusto/data/request.py
mordajon: Not sure what do you mean by this

azure-kusto-data/azure/kusto/data/request.py
mordajon: I prefered to duplicate this specific method, so the typing will be more accurate and informative

azure-kusto-data/azure/kusto/data/security.py
mordajon: I prefer that the logic of expiration will be written once... and not twice

mordajon:
ExecuteRequestParams should take authorization as param in addition to self._request_headers

.travis.yml
```diff
@@ -1,23 +1,24 @@
...
-  fast_finish: true
+os:
+  - linux
+env:
+  - OPTIONAL=None
+  - OPTIONAL=aio
+  - OPTIONAL=pandas
+python:
+  - "3.5"
+  - "3.6"
```
danield137: 🎉

azure-kusto-data/azure/kusto/data/security.py
danield137: you can have a ctor that accepts the helper itself..

azure-kusto-data/azure/kusto/data/security.py
danield137: if you want to, then you need to promote time to be a parameter, with a default value like:
`is_expired(expiration_date, expiry_time=1 * 60`

I'm sure you're familiar with python's zen (explicit is better than implicit)

roeap:
Really looking forward for this feature to become available. Is there a plan to merge this PR soon?

woanware:
Have been testing the PR and it works in my scenario's, would be great to get it merged!

woanware:
Since using the PR code (under Windows 10, Python 3.7.6 x64) I have noticed regular asyncio errors, when running multiple async queries, for example:

```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
C:\Python37\lib\runpy.py in _run_module_as_main(mod_name, alter_argv)
    191         sys.argv[0] = mod_spec.origin
    192     return _run_code(code, main_globals, None,
--> 193                      "__main__", mod_spec)
    194 
    195 def run_module(mod_name, init_globals=None,

C:\Python37\lib\runpy.py in _run_code(code, run_globals, init_globals, mod_name, mod_spec, pkg_name, script_name)
     83                        __package__ = pkg_name,
     84                        __spec__ = mod_spec)
---> 85     exec(code, run_globals)
     86     return run_globals
     87 

D:\Dev\venv\lib\site-packages\ipykernel_launcher.py in <module>
     14 
     15     from ipykernel import kernelapp as app
---> 16     app.launch_new_instance()

D:\Dev\venv\lib\site-packages\traitlets\config\application.py in launch_instance(cls, argv, **kwargs)
    662         app = cls.instance(**kwargs)
    663         app.initialize(argv)
--> 664         app.start()
    665 
    666 #-----------------------------------------------------------------------------

D:\Dev\venv\lib\site-packages\ipykernel\kernelapp.py in start(self)
    581         self.io_loop = ioloop.IOLoop.current()
    582         try:
--> 583             self.io_loop.start()
    584         except KeyboardInterrupt:
    585             pass

D:\Dev\venv\lib\site-packages\tornado\platform\asyncio.py in start(self)
    130             self._setup_logging()
    131             asyncio.set_event_loop(self.asyncio_loop)
--> 132             self.asyncio_loop.run_forever()
    133         finally:
    134             asyncio.set_event_loop(old_loop)

C:\Python37\lib\asyncio\base_events.py in run_forever(self)
    536             events._set_running_loop(self)
    537             while True:
--> 538                 self._run_once()
    539                 if self._stopping:
    540                     break

C:\Python37\lib\asyncio\base_events.py in _run_once(self)
   1744                            timeout * 1e3, dt * 1e3)
   1745         else:
-> 1746             event_list = self._selector.select(timeout)
   1747         self._process_events(event_list)
   1748 

C:\Python37\lib\selectors.py in select(self, timeout)
    321         ready = []
    322         try:
--> 323             r, w, _ = self._select(self._readers, self._writers, [], timeout)
    324         except InterruptedError:
    325             return ready

C:\Python37\lib\selectors.py in _select(self, r, w, _, timeout)
    312     if sys.platform == 'win32':
    313         def _select(self, r, w, _, timeout=None):
--> 314             r, w, x = select.select(r, w, w, timeout)
    315             return r, w + x, []
    316     else:

ValueError: too many file descriptors in select()
```

woanware:
To fix the "too many file selectors" error I had to change the following line, since it kept the HTTP socket open:

```
```

To:

```
self._request_headers = {'Connection':'close', "Accept": "application/json", "Accept-Encoding": "gzip,deflate", "x-ms-client-version": "Kusto.Python.Client:" + VERSION}
```

woanware:
Bump?

vladikbr:
Already committed

PR #203, opened 2020-02-09
Fix MSI Auth

yogilad:
MSI tokens are using OAuth2 format while code seems to work using a different format.
Fixed _AadHelper._get_header_from_dict() to implicitly support OAuth2 token formats

### DIFF
--- a/azure-kusto-data/azure/kusto/data/security.py
+++ b/azure-kusto-data/azure/kusto/data/security.py
@@ -8,7 +8,7 @@ from urllib.parse import urlparse
 
 import dateutil.parser
 from adal import AuthenticationContext, AdalError
-from adal.constants import TokenResponseFields, OAuth2DeviceCodeResponseParameters
+from adal.constants import TokenResponseFields, OAuth2DeviceCodeResponseParameters, OAuth2ResponseParameters
 from msrestazure.azure_active_directory import MSIAuthentication
 
 from .exceptions import KustoClientError, KustoAuthenticationError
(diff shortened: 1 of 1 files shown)


azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -212,7 +212,13 @@ def get_token_from_msi(self) -> dict:
...
 
 def _get_header_from_dict(token: dict):
-    return _get_header(token[TokenResponseFields.TOKEN_TYPE], token[TokenResponseFields.ACCESS_TOKEN])
+    if TokenResponseFields.TOKEN_TYPE in token:
+        return _get_header(token[TokenResponseFields.TOKEN_TYPE], token[TokenResponseFields.ACCESS_TOKEN])
+    elif OAuth2ResponseParameters.TOKEN_TYPE in token:
+        # Assume OAuth2 format (e.g. MSI Token)
+        return _get_header(token[OAuth2ResponseParameters.TOKEN_TYPE], token[OAuth2ResponseParameters.ACCESS_TOKEN])
+    else:
+        raise KustoClientError("Unexpected token format!")
```
danield137: hmm, call to action? what should the client do in this case?

yogilad: hmm,
I did not want to leave the OAuth2 format as an 'else' option.
The exception is collected and re-thrown to the user.

Do you mean that the exception description is not detailed enough?
I can change it to "Unexpected token format. Unable to locate token type property."

danield137: I think the text is too technical. Thinking from the users's perspective - how will he know what to do? Not sure the user will get that this is an auth problem. perhaps have a `KustoAuthError` with some extra details regarding why we couldn't locate the token?

yogilad: There's nothing he can do but contact the developers (unless he is somehow constructing the token by hand).
That said, I'll be more specific regarding the missing jwt property names.

yogilad: P.S. this exception is caught and converted to KustoAuthenticationError by acquire_authorization_header()

PR #205, opened 2020-03-11
fix user size override

danield137:


azure-kusto-ingest/azure/kusto/ingest/descriptors.py
```diff
@@ -36,11 +36,12 @@ def __init__(self, path, size=0, source_id=None):
...
 
-        if self.path.endswith(".gz") or self.path.endswith(".zip"):
-            # TODO: this can be improved by reading last 4 bytes
-            self.size = int(os.path.getsize(self.path)) * 11
-        elif not self.size or self.size <= 0:
-            self.size = int(os.path.getsize(self.path))
+        if size == 0:
+            raw_size = int(os.path.getsize(self.path))
+            if self.path.endswith(".gz") or self.path.endswith(".zip"):
+                # TODO: this can be improved by reading last 4 bytes
```
yogilad: This comment is only relevant for .gz format.

danield137: yeah it's here as a reminder. let's talk about a better solution

PR #210, opened 2020-03-30
Auto-detect uncompressed size when ingesting Zip and gZip  files

yogilad:
Read gzip size from stream end
Read zip size from compressed archive metadata


azure-kusto-ingest/azure/kusto/ingest/descriptors.py
```diff
@@ -20,6 +22,10 @@ def assert_uuid4(maybe_uuid, error_message):
 class FileDescriptor:
     """FileDescriptor is used to describe a file that will be used as an ingestion source."""
 
+    # The compressed sized of a 4GB file that will not cause the gzip stream size (unsigned int) to rollover given a 1:40 compression factor
```
danield137: not sure I understand this. could you please try and rephrase the comment? I understand there is a max size for a single gzip file to store it's uncompressed size, but I don't understand what it is, and why we are assuming a 1:40 compression rate

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
```diff
@@ -20,6 +22,10 @@ def assert_uuid4(maybe_uuid, error_message):
...
+    c_gzip_max_disk_size_for_detections = int(4 * 1024 * 1024 * 1024 / 40)
+    c_default_compression_factor = 11
```
danield137: what does the c_ prefeix stand for?

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
```diff
@@ -37,11 +43,31 @@ def __init__(self, path, size=0, source_id=None):
         self.stream_name = os.path.basename(self.path)
 
         if size == 0:
-            raw_size = int(os.path.getsize(self.path))
-            if self.path.endswith(".gz") or self.path.endswith(".zip"):
-                # TODO: this can be improved by reading last 4 bytes
-                raw_size *= 11
-            self.size = raw_size
+            self._detect_size()
```
danield137: we try to avoid doing heavy lifting in ctors... let move this closer to usage (or hide it under a property function like .size)

azure-kusto-ingest/tests/test_descriptors.py
```diff
@@ -7,12 +7,15 @@
 class DescriptorsTest(unittest.TestCase):
     """Test class for FileDescriptor and BlobDescriptor."""
 
+    _uncompressed_size = 1569
```
danield137: there really is no need to hide the properties.. it is a test..

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
yogilad: Const.
I know python does not have such a concept, but figured this would make one weary of making changes to the value.

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
yogilad: Let me know what you think 

    # Gzip keeps the decompressed stream size as a UINT32 in the last 4 bytes of the stream, however this poses a limit to the expressed size which is 4GB
    # The standard says that when the size is bigger then 4GB, the UINT rolls over.
    # The below constant expresses the maximal size of a compressed stream that will not cause the UINT32 to rollover given a maximal compression ratio of 1:40

danield137: much better :) thanks!

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
danield137: the convention is just upper case

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
yogilad: Moved to property

PR #211, opened 2020-03-30
Undefined name: import io for lines 167 and 174

cclauss:
Undefined names have the potential to raise NameError at runtime.


danield137:
Hi @cclauss, good catch! thank you.

PR #234, opened 2020-05-14
Restore compatibility with dateutil 2.7.x

bluca:
#### Pull Request Description

Many distributions are using dateutil 2.7.x, including Debian stable.
Restore compatibility by truncating timestamps with a precision of
more than a microsecond.

---

#### Future Release Comment

**Fixes:**
- Restore compatibility with dateutil 2.7.x

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_converters.py
+++ b/azure-kusto-data/azure/kusto/data/_converters.py
@@ -7,13 +7,21 @@ from dateutil import parser
 
 # Regex for TimeSpan
 _TIMESPAN_PATTERN = re.compile(r"(-?)((?P<d>[0-9]*).)?(?P<h>[0-9]{2}):(?P<m>[0-9]{2}):(?P<s>[0-9]{2}(\.[0-9]+)?$)")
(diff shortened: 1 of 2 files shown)


danield137:
If I remember correctly the reason we dropped it is because it's not just precision loss (rounding the last number) it just dropped the last. This has the potential of producing wrong values without warning, and users need to be very aware of that. We tried writing code around the problem but decided it's not worth the effort.

I don't feel comfortable with going back to dropping the last digit because I don't feel python users should get a degraded version of the Kusto service.

Plus, since current tests take latest version, they don't really test the scenario of date-util==2.7.x
which I think will fail tests.

To be honest, to resolve this in a way that will not do the wrong math, is too much work right now, so I don't think it's worth the effort.

As a general note,
Looking at https://github.com/dateutil/dateutil/issues/653 suggests they are planing on jumping two major versions in the near feature and we will follow them.

**That being said, do feel free to open an issue** , and if we see it is a recurring ask, we will definitely invest the time in solving this

bluca:
Thanks for looking at this. From my POV, reduced functionality is better than no functionality at all. I'll just carry this patch as out-of-tree in Debian/Ubuntu.

PR #241, opened 2020-06-01
Fix missing exception message in KustoServiceError exception class

yogilad:
#### Pull Request Description

Fix error in Exception classes inheritance due to breaking changes in Python 3

---

#### Future Release Comment

**Fixes:**
- Fix a bug where the exception text description reported to KustoServiceError were not saved due to a change in the syntax for explicitly calling a Super class constructor. Please mind that,
-- The exception text description is saved in the `args` field of the `Exception` class. 
-- The `message` field no long exists in Python 3

### DIFF
--- a/azure-kusto-data/azure/kusto/data/exceptions.py
+++ b/azure-kusto-data/azure/kusto/data/exceptions.py
@@ -11,7 +11,7 @@ class KustoError(Exception):
 class KustoServiceError(KustoError):
     """Raised when the Kusto service was unable to process a request."""
 
-    def __init__(self, messages, http_response, kusto_response=None):
+    def __init__(self, messages: Union[str, List[dict]], http_response: requests.Response, kusto_response=None):
         super().__init__(messages)
         self.http_response = http_response
         self.kusto_response = kusto_response
--- a/azure-kusto-data/azure/kusto/data/exceptions.py
+++ b/azure-kusto-data/azure/kusto/data/exceptions.py
@@ -11,7 +11,7 @@ class KustoError(Exception):
 class KustoServiceError(KustoError):
     """Raised when the Kusto service was unable to process a request."""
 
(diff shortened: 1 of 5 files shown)


azure-kusto-data/azure/kusto/data/exceptions.py
```diff
@@ -9,8 +11,8 @@ class KustoError(Exception):
 class KustoServiceError(KustoError):
     """Raised when the Kusto service was unable to process a request."""
 
-    def __init__(self, messages, http_response, kusto_response=None):
-        super(KustoServiceError, self).__init__(self, messages)
+    def __init__(self, messages: Union[str, List[dict]], http_response: requests.Response, kusto_response=None):
```
ArielYehezkely, line 14: Using types is inconsistent with the rest of the code, I don't want to get into discussion whether its good or not to use types in Python, but I think we should be consistent.

yogilad: I think we should adopt a policy to modify files we touch to include types.
If that's ok with you, I can add types to the rest of the APIs in Exceptions (and any future file I modify).

ArielYehezkely: I prefer to keep it without types, we can have a meeting to discuss this.

yogilad: I removed the type hints.
Let's get this pushed and discuss the subject separately

azure-kusto-data/azure/kusto/data/exceptions.py
```diff
@@ -1,5 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
+from typing import List, Union
```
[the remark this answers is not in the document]
yogilad: Opted to implement for the other initializer instead

PR #242, opened 2020-06-02
Exact version dep on azure-kusto-data

danield137:
Fix dependency on `azure-kusto-data`


danield137:
more work needed before going ahead with this

PR #250, opened 2020-07-08
Rasie exception when az cli token not found - Closes #240

### LINKED ISSUE #240: Unhandled 'None' when failing to get Azure CLI token
#### Code Sample, a copy-pastable example if possible

```python
client = KustoClient(KustoConnectionStringBuilder.with_az_cli_authentication(CLUSTER_NAME))
# Some query
```
#### Problem description

In case an Azure CLI token is not available, the following exception is raised, which is not very informative:
```
File "azure\kusto\data\security.py", line 179, in _acquire_authorization_header
    TokenResponseFields.REFRESH_TOKEN in stored_token
TypeError: argument of type 'NoneType' is not iterable
```
The problem is the method `_get_azure_cli_auth_token` just returns `None`, and then the calling code does not handle it properly.

**Expected behavior**: Either fallback to a different authentication flow, or raise an exception with an informative error message.

#### If query related, does it happen on other platforms (Kusto Web UI, Kusto Explorer)? 

Not query related

#### Output of ``pip freeze`` 

<details>

adal==1.2.2
astor==0.8.1
astroid==2.3.3
atomicwrites==1.3.0
attrs==19.3.0
azure-kusto-data==0.1.0
certifi==2019.11.28
cffi==1.13.2
chardet==3.0.4
colorama==0.4.3
coverage==5.0.4
cryptography==2.8
entrypoints==0.3
flake8==3.7.9
flynt==0.46
idna==2.8
importlib-metadata==1.5.0
isodate==0.6.0
isort==4.3.21
lazy-object-proxy==1.4.3
mccabe==0.6.1
more-itertools==8.2.0
msrest==0.6.13
msrestazure==0.6.3
numpy==1.18.2
oauthlib==3.1.0
packaging==20.3
pandas==1.0.4
pluggy==0.13.1
py==1.8.1
pycodestyle==2.5.0
pycparser==2.19
pyflakes==2.1.1
PyJWT==1.7.1
pylint==2.4.4
pyparsing==2.4.6
pytest==5.4.1
pytest-cov==2.8.1
python-dateutil==2.8.1
pytz==2019.3
requests==2.22.0
requests-oauthlib==1.3.0
six==1.13.0
typed-ast==1.4.1
urllib3==1.25.7
wcwidth==0.1.8
wrapt==1.11.2
zipp==3.1.0

</details>

ArielYehezkely:
#### Pull Request Description

Raise descriptive exception when az cli token not found when using az cli authentication.
---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Raise descriptive exception when az cli token not found when using az cli authentication.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/security.py
+++ b/azure-kusto-data/azure/kusto/data/security.py
@@ -48,9 +48,7 @@ def _get_azure_cli_auth_token() -> dict:
             # TODO: not sure I should take the first
             return data[0]
         except Exception as e:
-            raise KustoAuthenticationError(
-                AuthenticationMethod.az_cli_profile, e, message="Azure cli token was not found. Please run 'az login' to setup account."
-            )
+            raise KustoClientError("Azure cli token was not found. Please run 'az login' to setup account.", e)
 
 
 @unique
@@ -152,6 +150,8 @@ class _AadHelper:
                 kwargs["authority"] = AuthenticationMethod.token_provider.value
             elif self.auth_context is not None:
                 kwargs["authority"] = self.auth_context.authority.url
+            elif self.authentication_method is AuthenticationMethod.az_cli_profile:
+                kwargs["authority"] = AuthenticationMethod.az_cli_profile.value
 
             raise KustoAuthenticationError(self.authentication_method.value, error, **kwargs)
 
--- a/azure-kusto-data/azure/kusto/data/security.py
+++ b/azure-kusto-data/azure/kusto/data/security.py
@@ -47,8 +47,10 @@ def _get_azure_cli_auth_token() -> dict:
 
             # TODO: not sure I should take the first
             return data[0]
-        except Exception:
-            pass
+        except Exception as e:
+            raise KustoAuthenticationError(
+                AuthenticationMethod.az_cli_profile, e, message="Azure cli token was not found. Please run 'az login' to setup account."
+            )
 
 
 @unique
--- a/azure-kusto-data/azure/kusto/data/exceptions.py
+++ b/azure-kusto-data/azure/kusto/data/exceptions.py
@@ -43,9 +43,11 @@ class KustoAuthenticationError(KustoClientError):
     def __init__(self, authentication_method: str, exception: Exception, **kwargs):
         super().__init__()
         self.authentication_method = authentication_method
-        self.authority = kwargs["authority"]
-        self.kusto_cluster = kwargs["resource"]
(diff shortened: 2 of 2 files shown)


azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -47,8 +47,10 @@ def _get_azure_cli_auth_token() -> dict:
 
             # TODO: not sure I should take the first
             return data[0]
-        except Exception:
-            pass
+        except Exception as e:
+            raise KustoAuthenticationError(
+                AuthenticationMethod.az_cli_profile, e, message="Azure cli token was not found. Please run 'az login' to setup account."
+            )
```
yogilad, line 53: self.kusto_uri is the resource.
I know the exception is resource agnostic, but you can still fill it in.
You can also pass a mock authority, as done with MSI & Token Provider authentication

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -47,8 +47,10 @@ def _get_azure_cli_auth_token() -> dict:
 
             # TODO: not sure I should take the first
             return data[0]
-        except Exception:
-            pass
+        except Exception as e:
+            raise KustoAuthenticationError(
+                AuthenticationMethod.az_cli_profile, e, message="Azure cli token was not found. Please run 'az login' to setup account."
```
yogilad, line 52: This message is a bit specific for such a broad exception catch.
Would it fail here if az is not installed, not in env?
I'd consider a more broad message or more granular checks and messages.

ArielYehezkely: There are too many cases that can cause an exception here (file not found, empty file, wrong content, json parsing etc) - all of them should be solved by logging in to az cli.
We wrap the original exception, not replacing it, so this message its just a suggestion.

yogilad: Sure

azure-kusto-data/azure/kusto/data/security.py
ArielYehezkely: This function is a stand alone, no class context, and I don't think passing this as an argument worth it.

yogilad: You can throw KustoClientError instead and use the add the missing details in acquire_authorization_header()
The resource is already set, just add an authority as done for MSI and Token Provider

PR #255, opened 2020-07-22
Fix for msi auth

yogilad:
#### Pull Request Description

The azure.identity token result is different then msrestazure, causing msi auth to fail

---

#### Future Release Comment

**Fixes:**
- Fixed: MSI auth throws exception due to change in token result format

### DIFF
--- a/azure-kusto-data/azure/kusto/data/security.py
+++ b/azure-kusto-data/azure/kusto/data/security.py
@@ -178,7 +178,7 @@ class _AadHelper:
         # Obtain token from MSI endpoint
         if self.authentication_method == AuthenticationMethod.managed_service_identity:
             msi_token = self.get_token_from_msi()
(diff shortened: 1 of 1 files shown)


azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -10,6 +10,7 @@
 from adal import AuthenticationContext, AdalError
 from adal.constants import TokenResponseFields, OAuth2DeviceCodeResponseParameters, OAuth2ResponseParameters
 from azure.identity import ManagedIdentityCredential
+from azure.core.credentials import AccessToken
```
yogilad, line 13: @ArielYehezkely ,
Question: azure.core.credentials is a dependency of azure.identity which I use only to specific the return value of a function.
Can I avoid specifying it in setup.py 'install_requires'?

PR #261, opened 2020-08-03
parse decimal when using pandas

ArielYehezkely:
#### Pull Request Description

Fix parsing decimal values in KustoResultRow when using pandas DataFrame.

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Fix parsing decimal values in KustoResultRow when using pandas DataFrame.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/helpers.py
+++ b/azure-kusto-data/azure/kusto/data/helpers.py
@@ -33,6 +33,12 @@ def to_pandas_timedelta(raw_value, timedelta_value) -> "pandas.Timedelta":
     return pd.Timedelta(timedelta_value.total_seconds(), unit="ns")
 
 
+def to_decimal(raw_value, *args):
+    from decimal import Decimal
+
+    return Decimal(raw_value.strip(' "'))
+
+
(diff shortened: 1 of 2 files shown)


azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -33,6 +33,12 @@ def to_pandas_timedelta(raw_value, timedelta_value) -> "pandas.Timedelta":
     return pd.Timedelta(timedelta_value.total_seconds(), unit="ns")
 
 
+def to_decimal(raw_value, *args):
+    from decimal import Decimal
+
+    return Decimal(raw_value.strip(' "'))
```
yogilad, line 39: I would use lstrip & rstrip instead.
Otherwise this conversion may occur [ "1 2" => 12 ]

ArielYehezkely: Strip removes characters leading and trailing chars so it will give you the same behavior.

yogilad: Correct.
Sorry for that :)

PR #262, opened 2020-08-04
Fix few typos

ArielYehezkely:
#### Pull Request Description

Fix typos in ingest sample and ingestion properties backward compatibility section. 

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Fix typos in ingest sample and ingestion properties backward compatibility section.

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
@@ -163,7 +163,7 @@ class IngestionProperties:
         dataFormat = kwargs.get("dataFormat", None)
         ingestionMapping = kwargs.get("ingestionMapping", None)
         ingestionMappingType = kwargs.get("ingestionMappingType", None)
-        ingestionMappingReference = kwargs.get("ingestionMappingType", None)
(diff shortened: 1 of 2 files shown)


azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
```diff
@@ -163,7 +163,7 @@ def __init__(
         dataFormat = kwargs.get("dataFormat", None)
         ingestionMapping = kwargs.get("ingestionMapping", None)
         ingestionMappingType = kwargs.get("ingestionMappingType", None)
-        ingestionMappingReference = kwargs.get("ingestionMappingType", None)
+        ingestionMappingReference = kwargs.get("ingestionMappingReference", None)
```
yogilad, line 166: Isn't this a breaking change even if the old value was incorrect?

ArielYehezkely: Why would it be a breaking change? Its a bug in backward compatibility code..

yogilad: Oh, thought this was just a name change, no I see the same property name was used twice

PR #266, opened 2020-09-04
Support independent clouds URIs

ArielYehezkely:
#### Pull Request Description

Independent clouds has different endpoint suffix from public cloud.
Replace the hardcoded "core.windows.net" suffix with the actual URL suffix. 

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Support independent cloud endpoints URLs suffix.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/client.py
+++ b/azure-kusto-data/azure/kusto/data/client.py
@@ -326,7 +326,7 @@ class KustoConnectionStringBuilder:
     def with_aad_managed_service_identity_authentication(
         cls, connection_string: str, client_id: str = None, object_id: str = None, msi_res_id: str = None, timeout: int = None
     ) -> "KustoConnectionStringBuilder":
-        """"
+        """ "
         Creates a KustoConnection string builder that will authenticate with AAD application, using
         an application token obtained from a Microsoft Service Identity endpoint. An optional user
         assigned application ID can be added to the token.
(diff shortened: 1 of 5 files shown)


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -326,7 +326,7 @@ def with_az_cli_authentication(cls, connection_string: str) -> "KustoConnectionS
     def with_aad_managed_service_identity_authentication(
         cls, connection_string: str, client_id: str = None, object_id: str = None, msi_res_id: str = None, timeout: int = None
     ) -> "KustoConnectionStringBuilder":
-        """"
+        """ "
```
yogilad, line 329: there's an extra " here, no?

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -7,32 +7,33 @@
 from azure.kusto.data import KustoClient
 from azure.kusto.data._models import KustoResultTable
 
-_URI_FORMAT = re.compile("https://(\\w+).(queue|blob|table).core.windows.net/([\\w,-]+)\\?(.*)")
+_URI_FORMAT = re.compile("https://(\\w+).(queue|blob|table).(core.\\w+.net)/([\\w,-]+)\\?(.*)")
```
yogilad, line 10: for all uses of \w, are we sure dashes and perhaps other characters are not allowed?

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: yep 👍

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
ArielYehezkely: AFAIK other characters are not allowed, but I don't know all independent clouds suffixes..

yogilad (approved):

PR #267, opened 2020-09-10
Add wrong endpoint detection

yihezkel:
#### Pull Request Description

Customers commonly fail ingesting initially because they use the incorrect endpoint. Generally the Ingestion/DataManagement endpoint is the same as the Data/Engine endpoint, but prefixed with "ingest-".
We are adding detection of a non-ingestion endpoint when using an ingestion client.

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- If the ingest client is constructed with a non-DataManagement endpoint, the user will receive a KustoClientError with a suggestion of the most likely DataManagement endpoint, instead of a generic KustoServiceError.

**Fixes:**
- None

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_version.py
+++ b/azure-kusto-data/azure/kusto/data/_version.py
@@ -1,3 +1,3 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
-VERSION = "1.0.2"
+VERSION = "2.0.0"
--- a/azure-kusto-data/azure/kusto/data/_version.py
+++ b/azure-kusto-data/azure/kusto/data/_version.py
@@ -1,3 +1,3 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
-VERSION = "2.0.0"
+VERSION = "1.0.2"
--- a/azure-kusto-ingest/README.rst
+++ b/azure-kusto-ingest/README.rst
@@ -4,10 +4,10 @@ Microsoft Azure Kusto Ingest Library for Python
 .. code-block:: python
 
     from azure.kusto.data import KustoConnectionStringBuilder
-    from azure.kusto.ingest import KustoIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
+    from azure.kusto.ingest import QueuedIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
 
     ingestion_props = IngestionProperties(database="{database_name}", table="{table_name}", data_format=DataFormat.CSV)
-    client = KustoIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
+    client = QueuedIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
 
     file_descriptor = FileDescriptor("{filename}.csv", 15360)  # in this example, the raw (uncompressed) size of the data is 15KB (15360 bytes)
     client.ingest_from_file(file_descriptor, ingestion_properties=ingestion_props)
--- a/azure-kusto-ingest/README.rst
+++ b/azure-kusto-ingest/README.rst
@@ -4,10 +4,10 @@ Microsoft Azure Kusto Ingest Library for Python
 .. code-block:: python
 
     from azure.kusto.data import KustoConnectionStringBuilder
-    from azure.kusto.ingest import QueuedIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
+    from azure.kusto.ingest import KustoIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
 
     ingestion_props = IngestionProperties(database="{database_name}", table="{table_name}", data_format=DataFormat.CSV)
-    client = QueuedIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
+    client = KustoIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
 
     file_descriptor = FileDescriptor("{filename}.csv", 15360)  # in this example, the raw (uncompressed) size of the data is 15KB (15360 bytes)
     client.ingest_from_file(file_descriptor, ingestion_properties=ingestion_props)
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -136,8 +136,8 @@ class _ResourceManager:
         return self._authorization_context
 
     def retrieve_service_type(self):
-        command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
         try:
+            command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
             return command_result.primary_results[0][0][_SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
             return ""
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -9,6 +9,8 @@ from azure.kusto.data._models import KustoResultTable
 from azure.kusto.data.exceptions import KustoServiceError, KustoClientError
 
 _URI_FORMAT = re.compile("https://(\\w+).(queue|blob|table).(core.\\w+.\\w+)/([\\w,-]+)\\?(.*)")
+_SHOW_VERSION = ".show version"
+_SERVICE_TYPE_COLUMN_NAME = "ServiceType"
 
 
 class _ResourceUri:
@@ -64,9 +66,6 @@ class _IngestClientResources:
 
 
 class _ResourceManager:
-    _SHOW_VERSION = ".show version"
-    _SERVICE_TYPE_COLUMN_NAME = "ServiceType"
-
     def __init__(self, kusto_client: KustoClient):
         self._kusto_client = kusto_client
         self._refresh_period = timedelta(hours=1)
@@ -138,12 +137,12 @@ class _ResourceManager:
 
     def retrieve_service_type(self):
         try:
-            command_result = self._kusto_client.execute("NetDefaultDB", self._SHOW_VERSION)
+            command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
         except KustoServiceError:
-            raise KustoServiceError("Couldn't retrieve ServiceType because of a service exception executing {0}".format(self._SHOW_VERSION), None)
+            raise KustoServiceError("Couldn't retrieve ServiceType because of a service exception executing {0}".format(_SHOW_VERSION), None)
         except KustoClientError:
-            raise KustoClientError("Couldn't retrieve ServiceType because of a client exception executing {0}".format(self._SHOW_VERSION))
+            raise KustoClientError("Couldn't retrieve ServiceType because of a client exception executing {0}".format(_SHOW_VERSION))
         try:
-            return command_result.primary_results[0][0][self._SERVICE_TYPE_COLUMN_NAME]
+            return command_result.primary_results[0][0][_SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
-            raise KustoServiceError("Couldn't retrieve ServiceType because '{0}' didn't return any records".format(self._SHOW_VERSION))
+            raise KustoServiceError("Couldn't retrieve ServiceType because '{0}' didn't return any records".format(_SHOW_VERSION))
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -140,4 +140,4 @@ class _ResourceManager:
         try:
             return command_result.primary_results[0][0][_SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
-            raise KustoServiceError("Couldn't retrieve ServiceType because '{0}' didn't return any records".format(_SHOW_VERSION))
+            return ""
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -136,12 +136,7 @@ class _ResourceManager:
         return self._authorization_context
 
     def retrieve_service_type(self):
-        try:
-            command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
-        except KustoServiceError:
-            raise KustoServiceError("Couldn't retrieve ServiceType because of a service exception executing {0}".format(_SHOW_VERSION), None)
-        except KustoClientError:
-            raise KustoClientError("Couldn't retrieve ServiceType because of a client exception executing {0}".format(_SHOW_VERSION))
+        command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
         try:
             return command_result.primary_results[0][0][_SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -146,4 +146,4 @@ class _ResourceManager:
         try:
             return command_result.primary_results[0][0][self._SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
-            raise KustoServiceError("Couldn't retrieve ServiceType because '.show version' didn't return any records")
+            raise KustoServiceError("Couldn't retrieve ServiceType because '{0}' didn't return any records".format(self._SHOW_VERSION))
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -145,5 +145,5 @@ class _ResourceManager:
             raise KustoClientError("Couldn't retrieve ServiceType because of a client exception executing {0}".format(self._SHOW_VERSION))
         try:
             return command_result.primary_results[0][0][self._SERVICE_TYPE_COLUMN_NAME]
-        except:
+        except (TypeError, KeyError):
             raise KustoServiceError("Couldn't retrieve ServiceType because '.show version' didn't return any records")
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -6,6 +6,7 @@ from typing import List
 
 from azure.kusto.data import KustoClient
 from azure.kusto.data._models import KustoResultTable
+from azure.kusto.data.exceptions import KustoServiceError, KustoClientError
 
 _URI_FORMAT = re.compile("https://(\\w+).(queue|blob|table).(core.\\w+.\\w+)/([\\w,-]+)\\?(.*)")
 
@@ -63,6 +64,9 @@ class _IngestClientResources:
 
 
 class _ResourceManager:
+    _SHOW_VERSION = '.show version'
+    _SERVICE_TYPE_COLUMN_NAME = 'ServiceType'
+
     def __init__(self, kusto_client: KustoClient):
         self._kusto_client = kusto_client
         self._refresh_period = timedelta(hours=1)
@@ -131,3 +135,15 @@ class _ResourceManager:
     def get_authorization_context(self):
         self._refresh_authorization_context()
         return self._authorization_context
+
+    def retrieve_service_type(self):
+        try:
+            command_result = self._kusto_client.execute("NetDefaultDB", self._SHOW_VERSION)
+        except KustoServiceError:
+            raise KustoServiceError("Couldn't retrieve ServiceType because of a service exception executing {0}".format(self._SHOW_VERSION), None)
+        except KustoClientError:
+            raise KustoClientError("Couldn't retrieve ServiceType because of a client exception executing {0}".format(self._SHOW_VERSION))
+        try:
+            return command_result.primary_results[0][0][self._SERVICE_TYPE_COLUMN_NAME]
+        except:
+            raise KustoServiceError("Couldn't retrieve ServiceType because '.show version' didn't return any records")
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -64,8 +64,8 @@ class _IngestClientResources:
 
 
 class _ResourceManager:
-    _SHOW_VERSION = '.show version'
-    _SERVICE_TYPE_COLUMN_NAME = 'ServiceType'
+    _SHOW_VERSION = ".show version"
+    _SERVICE_TYPE_COLUMN_NAME = "ServiceType"
 
     def __init__(self, kusto_client: KustoClient):
         self._kusto_client = kusto_client
--- a/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
@@ -53,6 +53,6 @@ class KustoInvalidEndpointError(KustoClientError):
         message = "You are using '{0}' client type, but the provided endpoint is of ServiceType '{1}'. Initialize the client with the appropriate endpoint URI".format(
             expected_service_type, actual_service_type
         )
-        if not suggested_endpoint_url:
+        if suggested_endpoint_url:
             message = message + ": '" + suggested_endpoint_url + "'"
         super(KustoInvalidEndpointError, self).__init__(message)
--- a/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
@@ -44,3 +44,15 @@ class KustoMissingMappingReferenceError(KustoClientError):
     def __init__(self):
         message = "When stream format is json, mapping name must be provided."
         super(KustoMissingMappingReferenceError, self).__init__(message)
+
+
+class KustoInvalidEndpointError(KustoClientError):
+    """Raised when trying to ingest to invalid cluster type."""
+
+    def __init__(self, expected_service_type, actual_service_type, suggested_endpoint_url=None):
+        message = "You are using '{0}' client type, but the provided endpoint is of ServiceType '{1}'. Initialize the client with the appropriate endpoint URI".format(
+            expected_service_type, actual_service_type
+        )
+        if suggested_endpoint_url is not None:
+            message = message + ": '" + suggested_endpoint_url + "'"
+        super(KustoInvalidEndpointError, self).__init__(message)
--- a/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
@@ -53,6 +53,6 @@ class KustoInvalidEndpointError(KustoClientError):
         message = "You are using '{0}' client type, but the provided endpoint is of ServiceType '{1}'. Initialize the client with the appropriate endpoint URI".format(
             expected_service_type, actual_service_type
         )
-        if suggested_endpoint_url is not None:
+        if not suggested_endpoint_url:
             message = message + ": '" + suggested_endpoint_url + "'"
         super(KustoInvalidEndpointError, self).__init__(message)
--- a/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
@@ -133,15 +133,11 @@ class KustoIngestClient:
             self._endpoint_service_type = self._retrieve_service_type()
 
         if self._EXPECTED_SERVICE_TYPE != self._endpoint_service_type:
-            has_endpoint = True
             if self._suggested_endpoint_uri is None or not self._suggested_endpoint_uri.strip():
                 self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
                 if not self._suggested_endpoint_uri.strip():
-                    has_endpoint = False
-            if has_endpoint:
-                raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type, self._suggested_endpoint_uri)
-            else:
-                raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)
+                    raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)
+            raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type, self._suggested_endpoint_uri)
 
     def _retrieve_service_type(self):
         return self._resource_manager.retrieve_service_type()
--- a/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
@@ -129,15 +129,16 @@ class KustoIngestClient:
         queue_client.send_message(content=content)
 
     def _validate_endpoint_service_type(self):
-        if not self._endpoint_service_type:
-            self._endpoint_service_type = self._retrieve_service_type()
+        if not self._hostname_starts_with_ingest(self._connection_datasource):
+            if not self._endpoint_service_type:
+                self._endpoint_service_type = self._retrieve_service_type()
 
-        if self._EXPECTED_SERVICE_TYPE != self._endpoint_service_type:
-            if not self._suggested_endpoint_uri:
-                self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
+            if self._EXPECTED_SERVICE_TYPE != self._endpoint_service_type:
                 if not self._suggested_endpoint_uri:
-                    raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)
-            raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type, self._suggested_endpoint_uri)
+                    self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
+                    if not self._suggested_endpoint_uri:
+                        raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)
+                raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type, self._suggested_endpoint_uri)
 
     def _retrieve_service_type(self):
         return self._resource_manager.retrieve_service_type()
@@ -155,3 +156,10 @@ class KustoIngestClient:
                 #   "Couldn't generate suggested endpoint due to problem parsing datasource, with exception: {ex}. The correct endpoint is usually the Engine endpoint with '{self._INGEST_PREFIX}' prepended to the hostname."
                 pass
         return endpoint_uri_to_suggest_str
+
+    def _hostname_starts_with_ingest(self, datasource):
+        datasource_uri = urlparse(datasource)
+        hostname = datasource_uri.hostname
+        if hostname and hostname.startswith(self._INGEST_PREFIX):
+            return True
+        return False
--- a/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
@@ -26,9 +26,11 @@ class QueuedIngestClient:
     https://docs.microsoft.com/en-us/azure/data-explorer/ingest-data-overview#ingestion-methods
     """
 
-    _INGEST_PREFIX = 'ingest-'
-    _WRONG_ENDPOINT_MESSAGE = "You are using '{0}' client type, but the provided endpoint is of ServiceType '{1}'. Initialize the client with the appropriate endpoint URI"
-    _EXPECTED_SERVICE_TYPE = 'DataManagement'
+    _INGEST_PREFIX = "ingest-"
+    _WRONG_ENDPOINT_MESSAGE = (
+        "You are using '{0}' client type, but the provided endpoint is of ServiceType '{1}'. Initialize the client with the appropriate endpoint URI"
+    )
+    _EXPECTED_SERVICE_TYPE = "DataManagement"
 
     def __init__(self, kcsb: Union[str, KustoConnectionStringBuilder]):
         """Kusto Ingest Client constructor.
@@ -83,15 +85,14 @@ class QueuedIngestClient:
             descriptor = FileDescriptor(file_descriptor)
 
         should_compress = not (
-                ingestion_properties.format in [DataFormat.AVRO, DataFormat.ORC, DataFormat.PARQUET]
-                or descriptor.path.endswith(".gz")
-                or descriptor.path.endswith(".zip")
+            ingestion_properties.format in [DataFormat.AVRO, DataFormat.ORC, DataFormat.PARQUET]
+            or descriptor.path.endswith(".gz")
+            or descriptor.path.endswith(".zip")
         )
 
         with descriptor.open(should_compress) as stream:
             blob_name = "{db}__{table}__{guid}__{file}".format(
-                db=ingestion_properties.database, table=ingestion_properties.table,
-                guid=descriptor.source_id or uuid.uuid4(), file=descriptor.stream_name
+                db=ingestion_properties.database, table=ingestion_properties.table, guid=descriptor.source_id or uuid.uuid4(), file=descriptor.stream_name
             )
 
             random_container = random.choice(containers)
@@ -123,15 +124,15 @@ class QueuedIngestClient:
         queue_client.send_message(content=content)
 
     def validate_endpoint_service_type(self):
-        if not hasattr(self, '_endpoint_service_type') or len(self._endpoint_service_type) == 0 or self._endpoint_service_type.isspace():
+        if not hasattr(self, "_endpoint_service_type") or len(self._endpoint_service_type) == 0 or self._endpoint_service_type.isspace():
             self._endpoint_service_type = self._retrieve_service_type()
 
         if not self._EXPECTED_SERVICE_TYPE == self._endpoint_service_type:
             message = self._WRONG_ENDPOINT_MESSAGE.format(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)
-            if not hasattr(self, '_suggested_endpoint_uri') or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
+            if not hasattr(self, "_suggested_endpoint_uri") or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
                 self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
-            if not hasattr(self, '_suggested_endpoint_uri') or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
-                message += '.'
+            if not hasattr(self, "_suggested_endpoint_uri") or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
+                message += "."
             else:
                 message = "{0}: '{1}'".format(message, self._suggested_endpoint_uri)
             raise KustoClientError(message)
@@ -139,13 +140,13 @@ class QueuedIngestClient:
     def _retrieve_service_type(self):
         if self._resource_manager is not None:
             return self._resource_manager.retrieve_service_type()
-        return ''
+        return ""
 
     def _generate_endpoint_suggestion(self, datasource):
         """The default is not passing a suggestion to the exception String"""
-        endpoint_uri_to_suggest_str = ''
+        endpoint_uri_to_suggest_str = ""
         if not len(datasource) == 0 and not datasource.isspace():
-            endpoint_uri_to_suggest = urlparse(datasource) # Standardize URL formatting
-            endpoint_uri_to_suggest = urlparse(endpoint_uri_to_suggest.scheme + '://' + self._INGEST_PREFIX + endpoint_uri_to_suggest.hostname)
+            endpoint_uri_to_suggest = urlparse(datasource)  # Standardize URL formatting
+            endpoint_uri_to_suggest = urlparse(endpoint_uri_to_suggest.scheme + "://" + self._INGEST_PREFIX + endpoint_uri_to_suggest.hostname)
             endpoint_uri_to_suggest_str = endpoint_uri_to_suggest.geturl()
         return endpoint_uri_to_suggest_str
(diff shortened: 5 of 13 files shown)


azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
         content = ingestion_blob_info_json
         queue_client = queue_service.get_queue_client(queue=random_queue.object_name, message_encode_policy=TextBase64EncodePolicy())
         queue_client.send_message(content=content)
+
+    def validate_endpoint_service_type(self):
```
ArielYehezkely, line 129: "private"?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -17,18 +19,27 @@
...
+    )
+    _EXPECTED_SERVICE_TYPE = "DataManagement"
+
     def __init__(self, kcsb: Union[str, KustoConnectionStringBuilder]):
         """Kusto Ingest Client constructor.
         :param kcsb: The connection string to initialize KustoClient.
         """
         self._resource_manager = _ResourceManager(KustoClient(kcsb))
+        if not isinstance(kcsb, KustoConnectionStringBuilder):
+            kcsb = KustoConnectionStringBuilder(kcsb)
```
ArielYehezkely, line 41: If you create KCSB anyway, you can pass it to the KustoClient above and don't create it twice (inside KustoClient)

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
         content = ingestion_blob_info_json
         queue_client = queue_service.get_queue_client(queue=random_queue.object_name, message_encode_policy=TextBase64EncodePolicy())
         queue_client.send_message(content=content)
+
+    def validate_endpoint_service_type(self):
+        if not hasattr(self, "_endpoint_service_type") or len(self._endpoint_service_type) == 0 or self._endpoint_service_type.isspace():
```
ArielYehezkely, line 130: not self._endpoint_service_type.strip()

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+            self._endpoint_service_type = self._retrieve_service_type()
+
+        if not self._EXPECTED_SERVICE_TYPE == self._endpoint_service_type:
```
ArielYehezkely, line 133: !=

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+                message += "."
+            raise KustoClientError(message)
+
+    def _retrieve_service_type(self):
+        return self._resource_manager.retrieve_service_type()
+
+    def _generate_endpoint_suggestion(self, datasource):
+        """The default is not passing a suggestion to the exception String"""
+        endpoint_uri_to_suggest_str = ""
+        if not len(datasource) == 0 and not datasource.isspace():
```
ArielYehezkely, line 152: here as well, not datasource.strip()

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+    def validate_endpoint_service_type(self):
+        if not hasattr(self, "_endpoint_service_type") or len(self._endpoint_service_type) == 0 or self._endpoint_service_type.isspace():
+            self._endpoint_service_type = self._retrieve_service_type()
+
+        if not self._EXPECTED_SERVICE_TYPE == self._endpoint_service_type:
+            message = self._WRONG_ENDPOINT_MESSAGE.format(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)
+            has_endpoint = True
+            if not hasattr(self, "_suggested_endpoint_uri") or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
+                self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
+                if not hasattr(self, "_suggested_endpoint_uri") or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
```
ArielYehezkely, line 138: why double check? remove has_endpoint and check:
if self._suggested_endpoint_uri

azure-kusto-ingest/tests/test_kusto_ingest_client.py
```diff
@@ -21,6 +22,8 @@
 UUID_REGEX = "[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}"
 BLOB_NAME_REGEX = "database__table__" + UUID_REGEX + "__dataset.csv.gz"
 BLOB_URL_REGEX = "https://storageaccount.blob.core.windows.net/tempstorage/database__table__" + UUID_REGEX + "__dataset.csv.gz[?]sas"
+STORAGE_QUEUE_URL = "https://storageaccount.queue.core.windows.net/readyforaggregation-secured?sp=rl&st=2020-05-20T13:38:37Z&se=2020-05-21T13:38:37Z&sv=2019-10-10&sr=c&sig=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+TMP_STORAGE_QUEUE_URL = "https://storageaccount.blob.core.windows.net/tempstorage?sp=rl&st=2020-05-20T13:38:37Z&se=2020-05-21T13:38:37Z&sv=2019-10-10&sr=c&sig=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```
ArielYehezkely, line 26: TEMP_STORAGE_URL

azure-kusto-ingest/tests/test_kusto_ingest_client.py
```diff
@@ -156,6 +191,36 @@ def test_sanity_ingest_from_file(self, mock_uuid, mock_put_message_in_queue, moc
...
+        try:
+            ingest_client.ingest_from_file(file_path, ingestion_properties=ingestion_properties)
+        except Exception as e:
+            reached_exception = True
+            assert e.__class__ == KustoClientError
+            assert (
+                e.args[0]
+                == "You are using 'DataManagement' client type, but the provided endpoint is of ServiceType 'Engine'. Initialize the client with the appropriate endpoint URI: 'https://ingest-somecluster.kusto.windows.net'"
+            )
+        assert reached_exception is True
```
ArielYehezkely, line 222: You can use self.assertRaises

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -63,6 +64,9 @@ def is_applicable(self):
 
 
 class _ResourceManager:
+    _SHOW_VERSION = ".show version"
```
yogilad, line 67: you can make those function locals.
No need to add them at the class level.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: @yihezkel , you should probably set this value in __init__ to a default (e.g. "" or None) to simplify this line.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
         content = ingestion_blob_info_json
         queue_client = queue_service.get_queue_client(queue=random_queue.object_name, message_encode_policy=TextBase64EncodePolicy())
         queue_client.send_message(content=content)
+
+    def validate_endpoint_service_type(self):
+        if not hasattr(self, "_endpoint_service_type") or len(self._endpoint_service_type) == 0 or self._endpoint_service_type.isspace():
+            self._endpoint_service_type = self._retrieve_service_type()
```
yogilad, line 131: The exceptions thrown by this API are meaningless to the SDK user. 
Suggest you catch and wrap with a user friendly notice

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: Since this is a fault flow which needs fixing by the user, I see no point in caching the message creation.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +125,32 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+                self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
+                if not hasattr(self, "_suggested_endpoint_uri") or len(self._suggested_endpoint_uri) == 0 or self._suggested_endpoint_uri.isspace():
+                    has_endpoint = False
+            if has_endpoint:
+                message = "{0}: '{1}'".format(message, self._suggested_endpoint_uri)
+            else:
+                message += "."
+            raise KustoClientError(message)
+
+    def _retrieve_service_type(self):
```
yogilad, line 146: Just call the underlying single line of code

azure-kusto-ingest/tests/test_kusto_ingest_client.py
```diff
@@ -156,6 +191,36 @@ def test_sanity_ingest_from_file(self, mock_uuid, mock_put_message_in_queue, moc
...
+                missing_path_parts.append(path_part)
+
+        file_path = os.path.join(current_dir, *missing_path_parts)
+
+        reached_exception = False
+        try:
+            ingest_client.ingest_from_file(file_path, ingestion_properties=ingestion_properties)
+        except Exception as e:
+            reached_exception = True
+            assert e.__class__ == KustoClientError
```
yogilad, line 217: isinstance()

azure-kusto-ingest/tests/test_kusto_ingest_client.py
```diff
@@ -156,6 +191,36 @@ def test_sanity_ingest_from_file(self, mock_uuid, mock_put_message_in_queue, moc
...
+            assert (
+                e.args[0]
+                == "You are using 'DataManagement' client type, but the provided endpoint is of ServiceType 'Engine'. Initialize the client with the appropriate endpoint URI: 'https://ingest-somecluster.kusto.windows.net'"
+            )
```
yogilad, line 221: newline after scope exit

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yogilad: Also, why not just catch KustoClientError explicitly

yogilad (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: Are you sure? This is my first Python change, so I looked this up. The guidance was that if the variable might not exist (which is the case here), to use hasattr().
And if I use strip and the variable doesn't exist, won't it throw an exception?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: Because this flow is slightly more complex than it seems. The invisible case is if _suggested_endpoint_uri is defined in the first place.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -17,18 +19,27 @@
 from .ingestion_properties import DataFormat, IngestionProperties
 
 
-class KustoIngestClient:
+class QueuedIngestClient:
```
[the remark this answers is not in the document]
yihezkel: How do I do that?

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
yihezkel: Agree it's unnecessary at this point, but I think this is a healthier design for reuse. I often see constants that tend to be reused  (like prefixes, queries, column names, error messages,, etc) defined multiple times, and I think that defining them to be accessible in the first place helps prevent that.

WDYT?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I preferred this approach to initializing to a default because it's a tradeoff between the extra checks on those 2 lines vs dirtying code that generally doesn't need to worry about any of these things. In most cases, the variables I use to keep track of whether in the past we've encountered the wrong endpoint are meaningless and therefore adds unnecessary complexity, both in the init code and while debugging.

I think we can achieve both my and your goals (simplicity both when the wrong endpoint is irrelevant, and on this particular line) by extracting this line to a function. WDYT?

yihezkel: "Suggest you catch and wrap with a user friendly notice"
Are you referring to the Error type (KustoClientError), or the message?
I assume the former, because I do include a user-friendly message.

In terms of my approach, I did this to copy the paradigm in the Java SDK where it always returns to the client either a Client or Server exception. I thought this SDK was the same way and therefore the preferred way to do it, but now that I look deeper I see there are a couple that were added. I'll change the error type.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: "Just call the underlying single line of code"
This is "hachana l'mazgan". With this approach, we can add the same feature to the Streaming Ingest Client without duplicating code. I took this approach in the Java SDK as well. The parent class will define the same signature, but the Streaming Client's implementation will be more complex. All the other functions I added to this file (with 1 small exception) can then be moved to the parent class, as the code's identical between the 2 clients.

yihezkel: "Since this is a fault flow which needs fixing by the user, I see no point in caching the message creation."
The message isn't cached. The _endpoint_service_type and _suggested_endpoint_uri are cached similar to the .NET SDK so that if the code using the SDK executes many calls and the client with the incorrect endpoint is a non-critical part of that (e.g. the critical flow they're doing involves a correctly-configured client, and so they won't catch the issue with the incorrectly-configured client right away), it won't require a slow ".show version" lookup and suggestion computation every time it *is* used.

I agree with you that it's not a big win since it's an error state they'll have to fix anyway, but I think it is better, and I also preferred consistency with the other SDKs. WDYT?

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yihezkel: I changed this code to use self.assertRaises anyway, but just to call out my approach for catching that Exception (so you don't think I'm an idiot - what I wrote here does indeed look stupid :-) ):
When I tested this code by integrating in sample.py, its exception handling behavior was unexpected: If the integrating code used a try/except with anything other than "Exception" itself, the execution would stop after the exception was raised and reached the integrating code. But if I wanted it to continue executing after the "except", I had to specifically catch "Exception". I looked it up, and someone on StackOverflow came to that conclusion as well.

But I still may be an idiot, because I might've been wrong and instead there might've been another issue at play :-)

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yihezkel: I don't know the Python conventions. Is there a specific guide/lint we're using? Because the lint I use says specifically to use 1 line there.

Here's the video capture: https://microsoft-my.sharepoint.com/:v:/p/yischoen/EVHw4mswbJZPppTJUhlcrZYBMRcfpIsltM9JyfijsldcEQ?e=MAZWt0

ohadbitt: We lint with black
pip install black

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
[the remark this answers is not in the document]
yihezkel: I see the version in _version.py

This change is only to the ingest package - I assume both packages' versions have to match, right?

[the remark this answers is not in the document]
yihezkel: Is our convention MAJOR.MINOR.PATCH?

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yihezkel: Weird, I used black on that file, and this was the result. Also, it seems GitHub blocks merging if it fails black, right?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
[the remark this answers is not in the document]
yogilad: Do we really want to raise a major version for this?
This seems like an overkill.
I would revert the change.

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
yogilad: 1. Since python is not really private, it's odd adding this to the class.
2. The use case for reusability is slim, and if it exists, I'd reuse the whole function, not the constant.
3. If I'm not mistaken, with python this would be defined once even if it is function local.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: 1. I disagree, you over-complicate. Just because python allows creating class members at any point, it does not mean this is a good practice.
2. The later. This message in these exception are meaningless to the SDK user

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: If you want to cache drop the hasattr syntax and create an empty class member in the initialization

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yogilad: try:
   do something 
except KustoClientError:
   this is ok
except exception: 
   this test has failed, but the exception is swallowed

azure-kusto-data/azure/kusto/data/_version.py
```diff
@@ -1,3 +1,3 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
-VERSION = "1.0.2"
+VERSION = "2.0.0"
```
yogilad, line 3: I think you should revert instead.
Let's discuss this with @ArielYehezkely

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -44,3 +44,15 @@ class KustoMissingMappingReferenceError(KustoClientError):
     def __init__(self):
         message = "When stream format is json, mapping name must be provided."
         super(KustoMissingMappingReferenceError, self).__init__(message)
+
+
+class KustoInvalidEndpointError(KustoClientError):
```
yogilad, line 49: This has a side effect where anyone who used to catch kusto client exceptions will now crash

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -126,30 +126,28 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
         queue_client = queue_service.get_queue_client(queue=random_queue.object_name, message_encode_policy=TextBase64EncodePolicy())
         queue_client.send_message(content=content)
 
-    def validate_endpoint_service_type(self):
+    def __validate_endpoint_service_type(self):
```
yogilad, line 129: did you mean to make this private (one underscore) or mangled (two underscores)?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: to be clear i refer to exceptions that might be thrown from retrieve_service_type.
The user has no clue you run .show version in this context.
Try catch it and raise a meaningful exception, with the source exception attached internally

yogilad (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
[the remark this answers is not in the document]
ArielYehezkely: A. I agree with Yochai
B. You should mention breaking changes in the description which is used afterwards to create the release notes.
C. After completing the PR and when we decide to release a new version, we bump the version of both packages, create a tag and release. 
D. Yes we follow the standard convention.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
ArielYehezkely: I agree its better to init the parameters before and avoid those checks.
Is it even reasonable to get a whitespace here? If not you can remove the split and check only empty string, but you use "or" so you shouldn't run into exception here because the first condition will fail the check.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
ArielYehezkely: But _generate_endpoint_suggestion returns an empty string if failed so anyway in line 138 you have this variable set.

azure-kusto-data/azure/kusto/data/_version.py
ArielYehezkely: And please don't bump version in the same PR.

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
ArielYehezkely: Are you sure? it inherits KustoClientError..

yogilad: Hmm... you may have a point :)

azure-kusto-data/azure/kusto/data/_version.py
yihezkel: @yogilad I'm reverting, and will keep this as a separate PR.
I'm watching/starred this repo, so I'll be notified of all PRs. When I see something that introduces a different breaking change, I'll merge this change as well.

@ArielYehezkely Is the rationale that a bump in version is irreversible, while would be relevant if the PR with the change is reverted?

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
yihezkel: 1. Python does have scoping though, so if it was defined only locally to the method it wouldn't be visible elsewhere. The only other option besides a refactor of a specialized file/class for all commands would be to **move it to the file level**. I do want this command to be reusable, so **I'll do that**, then.
I suspect it's the same rationale that was used for defining the _URI_FORMAT at the top of this same file - it's currently once used once in this file, but it can conceivably be used elsewhere, and it would be a shame to do anything but discourage duplicating it.
2. I don't agree - the very next use case is adding the feature to the Streaming client, where this same command will be necessary, but it'd have to be executed differently (can't reuse the method).
3. If it was function local, the value wouldn't exist until that function was called, which isn't guaranteed in valid use cases (see #2).

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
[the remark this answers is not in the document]
yihezkel: Thanks for the explanations @yogilad and @ArielYehezkely !

As I commented elsewhere, I'm reverting the breaking change and saving it for the next time a breaking change is more necessary.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I did, thanks Yochai!

<aside>You know, for a language that prides itself in being *explicit* and *simple*, a system of underscores to denote 4 unique scenarios seems a bit implicit and complex.</aside>

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: @ArielYehezkely 
Yes the way I wrote it, it wouldn't fail because of the "hasattr".
I was responding to your suggestion to replace that with "not self._endpoint_service_type.strip()", which would fail if it wasn't defined.

"Is it even reasonable to get a whitespace here?"
My coding style places an emphasis on resilience. Right now, it can't have whitespace, but perhaps it could in the future. This method is a black box that should produce the correct output given any reasonable input. It's similar to (I presume) why in our C# code base we prefer to use "string.IsNullOrWhiteSpace()" wherever possible.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: 1. You don't see the benefit of keeping the unnecessary wrong-endpoint-detection variables out of the happy path flow's code, or you do but don't think it's worthwhile creating a class instance variable on the fly?

2. I agree that the error in retrieve_service_type isn't intuitive to the user because they wouldn't even know that getting the endpoint's version is part of the flow. But I can't think of anything more user-friendly for this issue, other than for them to just know that something went wrong in the Kusto Client or on the Kusto Server. And they are told exactly that, because the error raised is a KustoClientError or KustoServiceError.
Either way, they'll have to look into the SDK deeper or open a ticket, and the message is a sort of internal attachment that will help them debug/us address the ticket.
Can you think of a better message or error type that would actually help the SDK user to *not* have to look deeper into the SDK code/open a ticket?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: Resolving, as this is a duplicate of the discussion above about pros/cons of defining each class instance variable in the init vs. keeping variables that are irrelevant to the happy path and only used in the wrong-endpoint-detection feature from dirtying up the rest of the class.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: Yes, line 137 can be shortened. I kept it there because I think the consistency among the 3 lines that do the same thing (133, 135, 137) is more maintainable (easier to read and understand).

But since you think it's better to remove the hasattr from the check in the final line I'll change it.

azure-kusto-ingest/README.rst
```diff
@@ -4,10 +4,10 @@ Microsoft Azure Kusto Ingest Library for Python
 .. code-block:: python
 
     from azure.kusto.data import KustoConnectionStringBuilder
-    from azure.kusto.ingest import KustoIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
+    from azure.kusto.ingest import QueuedIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
```
yogilad, line 7: leftover

azure-kusto-ingest/README.rst
```diff
@@ -4,10 +4,10 @@ Microsoft Azure Kusto Ingest Library for Python
 .. code-block:: python
 
     from azure.kusto.data import KustoConnectionStringBuilder
-    from azure.kusto.ingest import KustoIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
+    from azure.kusto.ingest import QueuedIngestClient, IngestionProperties, FileDescriptor, BlobDescriptor, DataFormat
 
     ingestion_props = IngestionProperties(database="{database_name}", table="{table_name}", data_format=DataFormat.CSV)
-    client = KustoIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
+    client = QueuedIngestClient("https://ingest-{cluster_name}.kusto.windows.net")
```
yogilad, line 10: leftover

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -6,28 +6,37 @@
...
 from ._resource_manager import _ResourceManager
 from .descriptors import BlobDescriptor, FileDescriptor
+from .exceptions import KustoInvalidEndpointError
 from .ingestion_properties import DataFormat, IngestionProperties
 
 
 class KustoIngestClient:
     """
-    Kusto ingest client provides methods to allow ingestion into kusto (ADX).
+    Kusto queued ingest client provides methods to allow ingestion into kusto (ADX).
```
yogilad, line 25:
```suggestion
    Kusto ingest client provides methods to allow queued ingestion into kusto (ADX).
```

azure-kusto-ingest/tests/e2e.py
```diff
@@ -108,19 +110,19 @@ def test_table_json_mapping_reference():
 
 
 def engine_kcsb_from_env() -> KustoConnectionStringBuilder:
-    engine_cs = os.environ.get("ENGINE_CONNECTION_STRING")
-    app_id = os.environ.get("APP_ID")
-    app_key = os.environ.get("APP_KEY")
-    auth_id = os.environ.get("AUTH_ID")
+    engine_cs = "https://yischoen.westcentralus.dev.kusto.windows.net"
```
yogilad, line 113: This is not something you should be committing.

azure-kusto-ingest/tests/e2e.py
```diff
@@ -108,19 +110,19 @@ def test_table_json_mapping_reference():
...
+    app_id = "ba80e9b4-4855-46d1-8ef4-6f32df342ad4"
+    app_key = "_QgA0h8F-uJQcXS60cqOtjN_7CnKXDU-6t"
```
yogilad, line 115: I'm surprised git allowed this

azure-kusto-ingest/tests/e2e.py
```diff
@@ -139,7 +141,7 @@ def get_file_path() -> str:
 
 
 # Init clients
-test_db = os.environ.get("TEST_DATABASE")
+test_db = "YihezkelTest"
```
yogilad, line 144: You should basically undo everything to this point

yogilad (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I intentionally left this particular one as is because it's more accurate regardless of what the class is called, but I also like how you wrote it.

azure-kusto-ingest/README.rst
yihezkel: Had to invalidate and clear IDE's cache in order for it to identify these, for some reason.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: @ArielYehezkely 
Sorry, I misunderstood - you were only suggesting replacing "len(self._endpoint_service_type) == 0 or self._endpoint_service_type.isspace()" with "not self._endpoint_service_type.strip()". I'll do that.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: I see you resolved '2'

As for '1', in my opinion you are abusing a capacity of Python to add class members at any point.
A class which has different attributes depending on which APIs are called is prone to faults, hard to comprehend, and hard to debug. 
Additionally, Should I want to use those attributes elsewhere I'll be forced to add those checks again and again.

Keep in mind that good code needs to be readable and maintainable, and sometimes this comes at the expanse of absolute efficiency.

As a compromise, **if you need to use this more then once**, you can wrap this in a property which checks for existence and calls the function.

yogilad (changes requested):
Getting there 👍

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -44,3 +44,15 @@ class KustoMissingMappingReferenceError(KustoClientError):
...
+    """Raised when trying to ingest to invalid cluster type."""
+
+    def __init__(self, expected_service_type, actual_service_type, suggested_endpoint_url=None):
+        message = "You are using '{0}' client type, but the provided endpoint is of ServiceType '{1}'. Initialize the client with the appropriate endpoint URI".format(
+            expected_service_type, actual_service_type
+        )
+        if suggested_endpoint_url is not None:
```
ArielYehezkely, line 56: if not suggested_endpoint_url:

azure-kusto-data/azure/kusto/data/_version.py
ArielYehezkely: 1. Correct you don't want to revert this unintentionally.
2. We don't release new version for every change, so we can have few commits between releases and its easier to track the version bumping as a separate commit. (v.1.0.0 .. A-B-C .. v1.0.1) we create a tag for each release and its nicer to attach it to the relevant commit and not to the code change itself.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I didn't change (2). The error type and error message are how I originally wrote them. I had asked if you has a better suggestion. However, per another comment I just wrote, I'm removing those lines altogether now.


Regarding (1), I see **you and @ArielYehezkely both agree and insist, so I'll change it**.

But for the record, I disagree and find the approach I wrote to be easier to debug, and more readable and maintainable. I put those variables only there because the feature is confined to only that section of code and I can't conceive of a scenario where they'd be used elsewhere.

As I mentioned, this is my first Python PR I've ever written, so I looked a lot up. And as far as I read, defining instance variables at runtime is very much considered good practice in scenarios like this by the Python community.

In fact, I think the way I did it here is common among popular Python modules. I found many... 1 example is the "http" module's server.py's _headers_buffer.

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -131,3 +134,15 @@ def get_ingestions_status_tables(self) -> List[_ResourceUri]:
     def get_authorization_context(self):
         self._refresh_authorization_context()
         return self._authorization_context
+
+    def retrieve_service_type(self):
+        try:
+            command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
+        except KustoServiceError:
+            raise KustoServiceError("Couldn't retrieve ServiceType because of a service exception executing {0}".format(_SHOW_VERSION), None)
```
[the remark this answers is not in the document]
yihezkel: I thought rethrowing a cleaner error would simplify and thus best help the SDK user, but perhaps it's best not to get involved. I'm removing these so that whatever error is thrown will be raised without my simplified explanations.

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
yihezkel: Much of what I did here was based on searching best practices, since this my first Python work. I wrote this particular line according to: https://stackoverflow.com/questions/3965104/not-none-test-in-python
It says that for readability, "is not None" is better.
Should I change it anyway?

yihezkel:
Note that I added something new:
I added to ingest_client.py a "warning but dont fail" in case the datasource can't be parsed, in the form of a print statement.

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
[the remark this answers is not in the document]
yogilad: You misunderstand,  
The clear error is for the SDK user.
The encapsulated error is for us to be able to assist him when he has a problem.
The user does not care about the fact that you implicitly called .show version, but we care very much about why it had failed.
Different audience, different needs.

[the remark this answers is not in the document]
yihezkel: No I understand that we can try to address different audiences, but the fact is that:
1) The SDK user will see both exceptions if we rethrow while including the original exception instead of swallowing it
2) The SDK user will have to follow the same next steps either way. They will have to either debug/look deeper into the code, or open a support ticket.
For both of these reasons, adding a simplified message is not necessary, not a big benefit, and complicates the review.

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yihezkel: According to my tests, it works similarly, but with 1 slight difference:
try:
  do something
except KustoClientError:
  this is ok, **but if we get here execution will not proceed to "next_step()"**
except exception:
  this test has failed, but the exception is swallowed, **and if we get here execution will proceed to "next_step()"**
next_step()

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
yihezkel: Unless you say otherwise, I'm leaving it as it seems the Python community prefers it.

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yihezkel: Unless you say otherwise, I'm leaving this as I wrote it because I think I did it as black requires.

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
ArielYehezkely: The reason to use "x is not None" over simple "not x" is some types are falsy, and sometimes you can get the wrong result if you don't compare explicitly. 
Here I think you'd want to use str falsy nature so if the string is empty there is no point to edit the message.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +127,37 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
         content = ingestion_blob_info_json
         queue_client = queue_service.get_queue_client(queue=random_queue.object_name, message_encode_policy=TextBase64EncodePolicy())
         queue_client.send_message(content=content)
+
+    def _validate_endpoint_service_type(self):
+        if self._endpoint_service_type is None or not self._endpoint_service_type.strip():
```
ArielYehezkely, line 132: Maybe my question was lost in this long discussion, so asking again: is it reasonable to get a whitespace here? 
I not, simplify this check (and all other places) to "if self._endpoint_service_type"

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +127,37 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+            self._endpoint_service_type = self._retrieve_service_type()
+
+        if self._EXPECTED_SERVICE_TYPE != self._endpoint_service_type:
+            has_endpoint = True
```
ArielYehezkely, line 136: I don't see my previous comment so adding a new one. 
You don't need this variable, just do it this way:
```
if not self._suggested_endpoint_uri:
	self._suggested_endpoint_uri = self._generate_endpoint_suggestion(self._connection_datasource)
	if not self._suggested_endpoint_uri:
		raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type)

raise KustoInvalidEndpointError(self._EXPECTED_SERVICE_TYPE, self._endpoint_service_type, self._suggested_endpoint_uri)
```

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -108,3 +127,37 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+    def _generate_endpoint_suggestion(self, datasource):
+        """The default is not passing a suggestion to the exception String"""
+        endpoint_uri_to_suggest_str = ""
+        if datasource.strip():
+            try:
+                endpoint_uri_to_suggest = urlparse(datasource)  # Standardize URL formatting
+                endpoint_uri_to_suggest = urlparse(endpoint_uri_to_suggest.scheme + "://" + self._INGEST_PREFIX + endpoint_uri_to_suggest.hostname)
+                endpoint_uri_to_suggest_str = endpoint_uri_to_suggest.geturl()
+            except Exception as ex:
+                print(
```
ArielYehezkely, line 158: I don't think you'd want to print here. I agree we lack of logging mechanism but printing is not the alternative, packages not usually print unless special cases (device auth for example). Just pass here.

ArielYehezkely (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I agree that it was awkward to print. But I searched the Python SDK and found 1 other instance where we print as a way of providing a warning.

I realize this isn't ideal, but I did it because I couldn't think of a better way, and I think this is better than nothing. Is there a better way of doing it?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: Yeah my response got buried, sorry about that. In https://github.com/Azure/azure-kusto-python/pull/267#discussion_r488602561 I wrote:
My coding style places an emphasis on resilience. Right now, it can't have whitespace, but perhaps it could in the future. This method is a black box that should produce the correct output given any reasonable input. It's similar to (I presume) why in our C# code base we prefer to use "string.IsNullOrWhiteSpace()" wherever possible.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I find my way more readable, but I know most people prefer your way because it's more elegant and some even find it more readable. I'll update this now.

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
yihezkel: You make a good point, thanks for explaining!

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
ArielYehezkely: I don't think this is helpful, but confusing. User gets the actual error, and if the endpoint suggestion failed somehow its a secondary issue. User has nothing to do with this information unlike other places.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
ArielYehezkely: A black box defined by its input/output - so if the "contract" of the method is to return correct endpoint or empty 
 or None it is it's responsibility to return the right result so whoever use this method won't have to check unexpected values everywhere. 
You can verify in the method if you are concerned about getting whitespace.

yihezkel: Exactly, except inputs can also be inputs from other method calls within the method, and not necessarily just its literal signature.

"You can verify in the method if you are concerned about getting whitespace."
That's what I do here.
I'm not sure I followed your point, because the way I'm reading it, it seems you agree with me.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: IMHO, the concept of a warning is a valid and critical one. Sometimes you don't want to fail, but you need to let your user know that something less-than-ideal happened. That's the case here.

If you think the message itself is confusing, we can improve that. But I wrote it like that to provide both some technical data in case they want to dig into the code/debug, and also a high-level suggestion for how they can proceed.

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yogilad: Then don't catch Exception class

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: I agree with @ArielYehezkely.
As an SDK, stdout and stderr are not yours to write to.

azure-kusto-ingest/tests/test_kusto_ingest_client.py
yihezkel: I don't - that's outdated code since Ariel mentioned the assertRaises

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yihezkel: I agree with the problem. I had that conundrum when I was writing this. The question is how to make the code as good as it can reasonably be:
1) Ideally, we can add a logging infrastructure for these 2 and any future warnings. The downside to this is that it's out-of-scope for this change and might take time.
2) We already print elsewhere in the Python SDK. Therefore we're already in violation. I suggest we be consistent with our warnings. In that case we can print in both places and add a Work Item to implement #1 in the future.

I think one of these 2 solutions are best.

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -131,3 +134,10 @@ def get_ingestions_status_tables(self) -> List[_ResourceUri]:
     def get_authorization_context(self):
         self._refresh_authorization_context()
         return self._authorization_context
+
+    def retrieve_service_type(self):
+        command_result = self._kusto_client.execute("NetDefaultDB", _SHOW_VERSION)
```
yogilad, line 139: Re-opening the issue,
The user cares not that you decided to implicitly run _.show version_
Should execute fail, he would not know what to do with the server exception.
You can choose to,

yogilad (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: This print is not as important as you make it to be.
If you can't generate the URI and you already throw a less specific exception, just let this warning go.

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
yogilad, line 139: Move this line inside the try block and we are good to go

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -155,3 +156,10 @@ def _generate_endpoint_suggestion(self, datasource):
                 #   "Couldn't generate suggested endpoint due to problem parsing datasource, with exception: {ex}. The correct endpoint is usually the Engine endpoint with '{self._INGEST_PREFIX}' prepended to the hostname."
                 pass
         return endpoint_uri_to_suggest_str
+
+    def _hostname_starts_with_ingest(self, datasource):
+        datasource_uri = urlparse(datasource)
+        hostname = datasource_uri.hostname
+        if hostname and hostname.startswith(self._INGEST_PREFIX):
```
ArielYehezkely, line 163: nit -  return hostname and hostname.startswith(self._INGEST_PREFIX)

yogilad (approved):

ArielYehezkely (approved):

PR #269, opened 2020-09-22
Clean up data (drop created table) after running E2E tests

yihezkel:
#### Pull Request Description

Clean up data (drop created table) after running E2E tests.

I remove a separate function that drops the table which was executed before starting the tests. Since the generated table name includes a second-based timestamp, this should only be necessary if the previous execution was *started* within 1 second of the current execution, including the time it took to create the table. As surprising as that might be (even if it failed immediately after creating the table and was then reexecuted), the function must've been added when this or some other situation that made this necessary was encountered. Adding randomness to the table's name should solve the same issue, but faster to reduce the E2E's already long runtime.

---

### DIFF
--- a/azure-kusto-ingest/tests/e2e.py
+++ b/azure-kusto-ingest/tests/e2e.py
@@ -3,6 +3,7 @@
 import datetime
 import io
 import os
+import random
 import sys
 import time
 import uuid
@@ -29,9 +30,6 @@ CLEAR_DB_CACHE = ".clear database cache streamingingestion schema"
 
 class TestData:
     """A class to define mappings to deft table."""
-
-    def __init__(self):
-        pass
 
     @staticmethod
     def test_table_csv_mappings():
@@ -126,8 +124,8 @@ def dm_kcsb_from_env() -> KustoConnectionStringBuilder:
     return KustoConnectionStringBuilder.with_aad_application_key_authentication(dm_cs, app_id, app_key, auth_id)
 
 
-def clean_previous_tests(engine_client, database, table):
-    engine_client.execute(database, ".drop table {0} ifexists".format(table))
+def teardown_module():
+    client.execute(test_db, ".drop table {} ifexists".format(test_table))
 
 
 def get_file_path() -> str:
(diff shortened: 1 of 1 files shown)


azure-kusto-ingest/tests/e2e.py
```diff
@@ -144,14 +142,13 @@ def get_file_path() -> str:
 test_db = os.environ.get("TEST_DATABASE")
 
 python_version = "_".join([str(v) for v in sys.version_info[:3]])
-test_table = "python_test_{0}_{1}".format(python_version, str(int(time.time())))
+test_table = "python_test_{0}_{1}_{2}".format(python_version, str(int(time.time())), random.randint(1, 100000))
```
yogilad, line 145: GUID instead of random?

azure-kusto-ingest/tests/e2e.py
```diff
@@ -126,8 +124,8 @@ def dm_kcsb_from_env() -> KustoConnectionStringBuilder:
     return KustoConnectionStringBuilder.with_aad_application_key_authentication(dm_cs, app_id, app_key, auth_id)
 
 
-def clean_previous_tests(engine_client, database, table):
-    engine_client.execute(database, ".drop table {0} ifexists".format(table))
+def teardown_module():
```
yogilad, line 127: Is there some pytest convention that makes this API be called after the test is run?

yogilad (changes requested):

azure-kusto-ingest/tests/e2e.py
yihezkel: Exactly: https://docs.pytest.org/en/stable/xunit_setup.html

azure-kusto-ingest/tests/e2e.py
yihezkel: I used random because I thought a UUID would be overkill. Aren't Python UUIDs 36 characters long? And wouldn't it take longer for Python to generate its 32 (not including dashes) random characters than to generate 5 random digits?

yihezkel: I also considered using a random int *instead* of the python version and time components, but I think the 1st two fields are helpful to identify what (and when) created rogue Kusto tables.

yogilad: Probably, but is sends a clearer intent.
optional comment, not a must

yogilad (approved):

PR #279, opened 2020-11-05
Support norequesttimeout ClientRequestProperty

yihezkel:
#### Pull Request Description
Support norequesttimeout ClientRequestProperty by setting the timeout to its maximum (1h30s) + client-server delta.

---

#### Future Release Comment
**Features:**
- Implement ClientRequestProperty norequesttimeout, which sets the request timeout to its maximum.

**Fixes:**
- Add client-server delta to all request timeouts.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/client.py
+++ b/azure-kusto-data/azure/kusto/data/client.py
@@ -566,8 +566,8 @@ class KustoClient:
     `execute_mgmt`: executes a KQL control command against the Kusto service.
     """
 
-    _mgmt_default_timeout = timedelta(hours=1, seconds=30)
-    _query_default_timeout = timedelta(minutes=4, seconds=30)
+    _mgmt_default_timeout = timedelta(hours=1)
+    _query_default_timeout = timedelta(minutes=4)
     _streaming_ingest_default_timeout = timedelta(minutes=10)
     _client_server_delta = timedelta(seconds=30)
 
--- a/azure-kusto-data/azure/kusto/data/client.py
+++ b/azure-kusto-data/azure/kusto/data/client.py
@@ -694,7 +694,7 @@ class KustoClient:
             if properties.get_option(ClientRequestProperties.no_request_timeout_option_name, False):
                 timeout = KustoClient._mgmt_default_timeout
             else:
-                timeout = (properties.get_option(ClientRequestProperties.request_timeout_option_name, timeout))
+                timeout = properties.get_option(ClientRequestProperties.request_timeout_option_name, timeout)
 
         timeout = (timeout or KustoClient._mgmt_default_timeout) + KustoClient._client_server_delta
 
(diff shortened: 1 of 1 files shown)


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -679,19 +681,22 @@ def _execute(self, endpoint: str, database: str, query: str, payload: io.IOBase,
...
+                timeout = KustoClient._mgmt_default_timeout
+            else:
+                timeout = properties.get_option(ClientRequestProperties.request_timeout_option_name, timeout)
 
-        if self._auth_provider:
-            request_headers["Authorization"] = self._auth_provider.acquire_authorization_header()
-
-        if properties:
-            timeout = properties.get_option(ClientRequestProperties.request_timeout_option_name, timeout)
+        timeout = (timeout or KustoClient._mgmt_default_timeout) + KustoClient._client_server_delta
```
ArielYehezkely, line 699: Why this "or" needed? (timeout should be set by now)

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -568,6 +569,7 @@ class KustoClient:
     _mgmt_default_timeout = timedelta(hours=1, seconds=30)
     _query_default_timeout = timedelta(minutes=4, seconds=30)
     _streaming_ingest_default_timeout = timedelta(minutes=10)
+    _client_server_delta = timedelta(seconds=30)
```
ArielYehezkely, line 572: I guess you can remove the 30 seconds delta from the other defaults.

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/client.py
yihezkel: In all existing flows that call this method, timeout is set to a value. The goal of the "or" is to account for the possibility that None will be passed, without a relevant ClientRequestProperties.

azure-kusto-data/azure/kusto/data/client.py
yihezkel: I used the Java client (src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java) as a reference. There, :30 is in addition to (1:30|4:30|10:00).

What's the desired behavior?

yihezkel: I looked up how it works in the .NET SDK (Src/Common/Kusto.Cloud.Platform/Utils/ServiceModelRequestTimeout.cs), and I see that the intent is as you indicated - 1 hour for ingest, 4 minutes for queries, 10 minutes for streaming, and everything + 30 seconds client-server delta. I'll fix this both here and in the Java SDK.

ArielYehezkely (approved):

PR #280, opened 2020-11-13
Support TCP Keep-Alives and expose the retry parameter

yihezkel:
#### Pull Request Description

Support TCP Keep-Alives and expose a retry parameter in the KustoClient init function

---

#### Future Release Comment
**Features:**
- Send TCP Keep-Alives at regular intervals so that connections for long-running queries aren't dropped by the Load Balancer Service
- Expose to SDK users an optional retry parameter in the KustoClient init function

### DIFF
--- a/azure-kusto-data/azure/kusto/data/client.py
+++ b/azure-kusto-data/azure/kusto/data/client.py
@@ -602,7 +602,9 @@ class KustoClient:
 
         # Create a session object for connection pooling
         self._session = requests.Session()
-        adapter = HTTPAdapterWithSocketOptions(socket_options=HTTPConnection.default_socket_options + self.compose_socket_options(), pool_maxsize=self._max_pool_size, max_retries=retries)
+        adapter = HTTPAdapterWithSocketOptions(
+            socket_options=HTTPConnection.default_socket_options + self.compose_socket_options(), pool_maxsize=self._max_pool_size, max_retries=retries
+        )
         self._session.mount("http://", adapter)
         self._session.mount("https://", adapter)
 
--- a/azure-kusto-data/azure/kusto/data/client.py
+++ b/azure-kusto-data/azure/kusto/data/client.py
@@ -629,7 +629,7 @@ class KustoClient:
     def compose_socket_options():
         # Sends TCP Keep-Alive after MAX_IDLE_SECONDS seconds of idleness, once every INTERVAL_SECONDS seconds, and closes the connection after MAX_FAILED_KEEPALIVES failed pings (e.g. 20 => 1:00:30)
         MAX_IDLE_SECONDS = 30
-        INTERVAL_SECONDS = 180
+        INTERVAL_SECONDS = 180 # Corresponds to Azure Load Balancer Service 4 minute timeout, with 1 minute of slack
         MAX_FAILED_KEEPALIVES = 20
 
         if (
@@ -651,13 +651,11 @@ class KustoClient:
             and hasattr(socket, "SOL_SOCKET")
             and hasattr(socket, "SO_KEEPALIVE")
             and hasattr(socket, "TCP_KEEPIDLE")
-            and hasattr(socket, "TCP_KEEPINTVL")
             and hasattr(socket, "TCP_KEEPCNT")
         ):
             return [
                 (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),
                 (socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, MAX_IDLE_SECONDS),
-                (socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, INTERVAL_SECONDS),
                 (socket.IPPROTO_TCP, socket.TCP_KEEPCNT, MAX_FAILED_KEEPALIVES),
             ]
         if sys.platform == "darwin" and hasattr(socket, "SOL_SOCKET") and hasattr(socket, "SO_KEEPALIVE") and hasattr(socket, "IPPROTO_TCP"):
--- a/azure-kusto-data/azure/kusto/data/client.py
+++ b/azure-kusto-data/azure/kusto/data/client.py
@@ -629,7 +629,7 @@ class KustoClient:
     def compose_socket_options():
         # Sends TCP Keep-Alive after MAX_IDLE_SECONDS seconds of idleness, once every INTERVAL_SECONDS seconds, and closes the connection after MAX_FAILED_KEEPALIVES failed pings (e.g. 20 => 1:00:30)
         MAX_IDLE_SECONDS = 30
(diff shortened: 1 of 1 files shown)


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -584,15 +599,56 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
...
+            and hasattr(socket, "TCP_KEEPINTVL")
+            and hasattr(socket, "TCP_KEEPCNT")
+        ):
+            return [
+                (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),
+                (socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, MAX_IDLE_SECONDS),
+                (socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, INTERVAL_SECONDS),
+                (socket.IPPROTO_TCP, socket.TCP_KEEPCNT, MAX_FAILED_KEEPALIVES),
+            ]
+        if (
```
ArielYehezkely, line 634: Isn't it the same as "linux"?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -584,15 +599,56 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
...
         self._query_endpoint = "{0}/v2/rest/query".format(kusto_cluster)
         self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(kusto_cluster)
         # notice that in this context, federated actually just stands for add auth, not aad federated auth (legacy code)
         self._auth_provider = _AadHelper(kcsb) if kcsb.aad_federated_security else None
         self._request_headers = {"Accept": "application/json", "Accept-Encoding": "gzip,deflate", "x-ms-client-version": "Kusto.Python.Client:" + VERSION}
 
+    @staticmethod
+    def compose_socket_options():
+        # Sends TCP Keep-Alive after MAX_IDLE_SECONDS seconds of idleness, once every INTERVAL_SECONDS seconds, and closes the connection after MAX_FAILED_KEEPALIVES failed pings (e.g. 20 => 1:00:30)
+        MAX_IDLE_SECONDS = 30
```
ArielYehezkely, line 616: What are the default values and why do we need to change them? If we can let's keep the default values and save the different OS's settings.

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -584,15 +599,56 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
...
+
+        if (
+            sys.platform == "linux"
+            and hasattr(socket, "SOL_SOCKET")
+            and hasattr(socket, "SO_KEEPALIVE")
+            and hasattr(socket, "TCP_KEEPIDLE")
+            and hasattr(socket, "TCP_KEEPINTVL")
+            and hasattr(socket, "TCP_KEEPCNT")
+        ):
+            return [
```
ArielYehezkely, line 628: This way you'll override some of the default socket options like (from urllib3): 
`#: Disable Nagle's algorithm by default.`
`#: [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]`
`default_socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]`

My suggestion:
`HTTPConnection.default_socket_options + [`
`                                (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),`
`                                ...`
`                            ]`

yihezkel: Good catch - I see that now in urllib3.connection.HTTPConnection. Thanks!

azure-kusto-data/azure/kusto/data/client.py
yihezkel: I'll research this further, but the problem is that it's not really documented. As far as I understand, we need to set all values, including the timeouts, for it to work (no default).

But even if there were defaults, I'd think it'd be best to set consistent Keep-Alive timings across all OSes, and deliberately for what we want. Why use an OS default?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -572,7 +587,7 @@ class KustoClient:
     # The maximum amount of connections to be able to operate in parallel
     _max_pool_size = 100
 
-    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
+    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str], retries=0):
```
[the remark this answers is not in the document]
yihezkel: - I agree that the name should be improved, if kept there.
- I would like to keep the socket options hardcoded, so I'd rather not combine that with setting the max_pool_size or retries.

Regarding the design:
Wouldn't it be helpful to explicitly expose this option to the SDK-user? (I thought I saw a request recently for exactly that.)
If not, why not remove it altogether? In other words, what's the point of exposing it at all if we're not exposing it to the SDK-user, and the value will be set to its default of 0 in requests.adapters.HTTPAdapter?

Or is your approach to compose the HTTPAdapter so the SDK-user can set its values as if it's an undocumented feature (so they can write something along the lines of `kusto_client_instance.adapter.http_retries = 3`)?

If this design is more than a quick chat, let's discuss on the phone.

azure-kusto-data/azure/kusto/data/client.py
yihezkel: Yes, but I want them to be separately tweakable. I'm also hoping to reduce what's needed for Windows, as I think some of them are unnecessary.

azure-kusto-data/azure/kusto/data/client.py
yihezkel: I researched and tested, and indeed there are no defaults for many of these. For example, on Windows if you just set:
`(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)`
But not:
`(socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, [number of seconds])`
Then no TCP Keep-Alives will be sent.

azure-kusto-data/azure/kusto/data/client.py
yihezkel: I researched and tested, and indeed the following is ignored by WIndows and is therefore unnecessary::
`(socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, INTERVAL_SECONDS)`
I'm removing it for Windows.

I tested Linux, and it respects both TCP_KEEPIDLE and TCP_KEEPINTVL.

azure-kusto-data/azure/kusto/data/client.py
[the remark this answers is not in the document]
yihezkel: Implemented as explicitly exposed, but via a setter method instead of constructor.

ArielYehezkely (approved):

PR #201, opened 2020-02-03
initial az cli support

danield137:
https://github.com/Azure/azure-kusto-python/issues/139

PR #288, opened 2020-12-15
Update docs for Azure CLI

ChrisCoe:
#### Pull Request Description

See #285

PR #285, opened 2020-12-10
Update Documentation to include Azure CLI auth method

ChrisCoe:
#### Pull Request Description

_Add more documentation to increase visibility for az cli authentication method for kusto client_

**Motive:** I tried several other authentication methods that did not fit my need before finding this method. Given the work for the implementation is there (#201), we should make it easier for others to find and benefit from.

PR #289, opened 2020-12-21
Update azure cli docs

ChrisCoe:
#### Pull Request Description

_Add documentation to increase visibility for az cli authentication method for kusto client_

**Motive:** I tried several other authentication methods that did not fit my need before finding this method. Given the work for the implementation is there (#201), we should make it easier for others to find and benefit from.


ChrisCoe:
@yogilad Let's push this PR to master with cleaner commit. This includes the exact same changes as #285 and #288.

README.md
```diff
@@ -43,6 +43,7 @@ Download a zip of the code via GitHub or PyPi. Then follow the same instructions
 * AAD Username/password - Provide your AAD username and password to Kusto client (**check the notice below**).
 * AAD application - Provide app ID and app secret to Kusto client.
 * AAD code - Provide only your AAD username, and authenticate yourself using a code, generated by ADAL.
+* AZ CLI - For those already using [azure-cli](https://github.com/Azure/azure-cli), provide access token by running `az account get-access-token`.
```
yogilad, line 46:
```suggestion
* AZ CLI - For those already using [azure-cli](https://github.com/Azure/azure-cli), provide access token for the logged in user`.
```

yogilad: The python impl. does not actually invoke az commands.
It inspects the az cache for the refresh token and uses it to generate tokens for other resources

azure-kusto-data/tests/sample.py
```diff
@@ -52,6 +52,10 @@
 user_assigned_client_id = "the AAD identity client id"
 kcsb = KustoConnectionStringBuilder.with_aad_managed_service_identity_authentication(cluster, client_id=user_assigned_client_id)
 
+# In case you want to authenticate with Azure CLI.
+# Running "az login" is not sufficient. Need to run "az account get-access-token" to cache token locally.
```
yogilad, line 56:
```suggestion
# Users are required to be in a logged in state in az-cli, for this authentication method to succeed. Run `az login` to login to azure cli.

```

yogilad: I actually checked this and az login if sufficient (commented in #285 IIRC)

yogilad (changes requested):

azure-kusto-data/tests/sample.py
ChrisCoe: Thanks for checking. Yes, I can clear this up a bit. Tokens have an expiration time of 60 minutes max. Az cli automatically refreshes tokens when needed for internal az cli use. Given azure-kusto-python is external to az cli and wants to use a token from there, we need to trigger a refresh to guarantee access. I found this info [here](https://github.com/Azure/azure-cli/issues/6234#issuecomment-387848841).

This also follows the contributor's [comments](https://github.com/Azure/azure-kusto-python/releases/tag/v0.0.43) for this feature. 

How about if we use this? Let me know.

Users are required to be in a logged in state in azure cli with an active access token, for this authentication method to succeed. Run `az login` to login to azure cli. Run `az account get-access-token` to refresh token if needed.

README.md
ChrisCoe: Good catch. Did not mean to imply the python implementation invokes az commands.

azure-kusto-data/tests/sample.py
yogilad: The way our az-cli integration works in kusto-python is that it picks up the cached refresh token and passes it to MSAL.
Once MSAL has that, it can produce new tokens and new refresh tokens on it's own (that's what the refresh token is there for).
If the initial attempt succeeded, the python SDK auth is now independent of az's.

ChrisCoe: Ah okay, I see now. I looked at the src code before but must have missed that. Thank you for filling my gap. Glad it was implemented that way as this is much cleaner:) Approving suggestion.

yogilad (approved):

ChrisCoe:
@yogilad Can someone with write access merge this?

yogilad:
@ChrisCoe , Done.
Thanks for the contribution 👍

PR #291, opened 2020-12-30
Lazy query results

ArielYehezkely:
#### Pull Request Description

Query results are converted from the raw Json objects into relational objects for easy iteration and data access. 
Currently the conversion is done row by row upon query response as there is a need to convert Kusto types to Python types what make calls to execute take longer than the actual http call time.

This change will hold the raw columns and rows, and will compute the Kusto table objects in a lazy manner.
The raw data could be used to convert the data to Pandas DataFrame in parallel and save the row by row conversion which is an unnecessary time consuming operation for Pandas users. 

#### Future Release Comment

Make query result conversion lazy and improve conversion into Pandas DataFrame.

**Breaking Changes:**
- None

**Features:**
- Improve calls to execute query performance by processing the result in a lazy manner. 
- Expose raw query result to improve conversion into Pandas DataFrame.

**Fixes:**
- None


azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -99,6 +99,11 @@ def __repr__(self):
         values = [repr(val) for val in self._value_by_name.values()]
         return "KustoResultRow(['{}'], [{}])".format("', '".join(self._value_by_name), ", ".join(values))
 
+    def __eq__(self, other):
+        if len(self) != len(other):
+            return False
+        return repr(self) == repr(other)
```
yogilad, line 105: This feels slow and with an overhead of allocations.
Why not compare value_by_name directly?

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -123,16 +128,29 @@ def __init__(self, json_table: dict):
...
             raise KustoServiceError(errors[0]["OneApiErrors"][0]["error"]["@message"], json_table)
 
-        self.rows = [KustoResultRow(self.columns, row) for row in json_table["Rows"]]
+        self.raw_columns = json_table["Columns"]
+        self.raw_rows = json_table["Rows"]
+        self.kusto_result_rows = None
+
+    @property
+    def rows(self):
+        if self.kusto_result_rows:
```
yogilad, line 137: If is None: 
  create 
return once

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -123,16 +128,29 @@ def __init__(self, json_table: dict):
...
+            return self.kusto_result_rows
+        self.kusto_result_rows = [KustoResultRow(self.columns, row) for row in self.raw_rows]
+        return self.kusto_result_rows
 
     @property
     def _rows(self) -> Iterator:
-        for row in self.rows:
-            yield row._hidden_values
+        for row_index, row in enumerate(self.raw_rows):
+            if self.kusto_result_rows:
```
yogilad, line 145: if is None... (same as above)

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -146,11 +164,17 @@ def __len__(self):
         return self.rows_count
 
     def __iter__(self):
-        for row in self.rows:
-            yield row
+        for row_index, row in enumerate(self.raw_rows):
+            if self.kusto_result_rows:
```
yogilad, line 168: You get the gist :)

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -146,11 +164,17 @@ def __len__(self):
...
+                yield self.kusto_result_rows[row_index]
+            else:
+                kusto_result_row = KustoResultRow(self.columns, row)
+                yield kusto_result_row
 
     def __getitem__(self, key):
-        return self.rows[key]
+        if self.kusto_result_rows:
```
yogilad, line 175: Shouldn't something be set if the first condition fails?

yogilad (changes requested):

azure-kusto-data/azure/kusto/data/_models.py
ArielYehezkely: Here its not the same actually. 
In rows - we'll convert all rows, here if we already have the converted rows, use them, otherwise convert row by row

azure-kusto-data/azure/kusto/data/_models.py
ArielYehezkely: No, __iter__ and __getitem__ shouldn't make changes

yogilad: wouldn't this cause problems for some use cases?

ArielYehezkely: I don't think so. Do I miss something?

yogilad: It seems like accessing the index operator without first "constructing" the result by some other API call, will create results ad-hoc, so if I were to iterate or access them multiple times, it would cause a lot of rows to be created and then disposed (mind my c#)

yogilad (approved):

PR #297, opened 2021-01-19
Logging 2

JF6:
#### Pull Request Description

If am fond of Kusto and also of the logging capabilities in Python... Hence a proposed logging handler to capture log and traces from the logging module directly in a Kusto table. 

Does only csv ingestion (I left the parameter for future improvements, but it can be removed)
Tested on both Windows & Linux. Works on both platforms, even if there are failures in pytest on Windows.

---

#### Future Release Comment
Requires versions 2.0.0 of both Data & Ingest + Pandas.

**Breaking Changes:**
- N/A

**Features:**
- Logging module integration.

**Fixes:**
- N/A


yogilad:
Hi @JF6 , we usually do not do this in the SDKs of Kusto.
We do plan on providing an Open Telemetry exporter in the future which would allow users to push logs out from any supporting language to Kusto.
In the meantime, you can use our Log Stash integration https://github.com/Azure/logstash-output-kusto

PR #299, opened 2021-01-21
Async io v2

AsafMah:
Adds async io options for our client, using aiohttp.

Re-made version of https://github.com/Azure/azure-kusto-python/pull/202
Since that PR we moved from ADAL to MSAL, so I re-wrote the relevant code.
Note that since most of our security options (except MSI) don't support native async, I went with a different method.

Also note that this drops support for python 3.5, as aiohttp doesn't support it either.


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -608,12 +667,11 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
         self._session.mount("http://", adapter)
         self._session.mount("https://", adapter)
 
-        self._mgmt_endpoint = "{0}/v1/rest/mgmt".format(kusto_cluster)
-        self._query_endpoint = "{0}/v2/rest/query".format(kusto_cluster)
-        self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(kusto_cluster)
+        self._mgmt_endpoint = "{0}/v1/rest/mgmt".format(self._kusto_cluster)
```
ArielYehezkely, line 670: why both here and base?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -726,67 +790,30 @@ def execute_streaming_ingest(
         if mapping_name is not None:
             endpoint = endpoint + "&mappingName=" + mapping_name
 
-        self._execute(endpoint, database, None, stream, KustoClient._streaming_ingest_default_timeout, properties)
+        try:
+            self._execute(endpoint, database, None, stream, self._streaming_ingest_default_timeout, properties)
+        except KustoRequestException as e:
+            raise KustoServiceError(
```
ArielYehezkely, line 796: In streaming ingest when the cluster is not configured for it, the call to ingest will return 404 -  this is why we've added the detailed error with relevant suggestion, please keep it.

azure-kusto-data/azure/kusto/data/exceptions.py
```diff
@@ -40,6 +40,15 @@ class KustoClientError(KustoError):
     """Raised when a Kusto client is unable to send or complete a request."""
 
 
+class KustoRequestException(KustoError):
```
ArielYehezkely, line 43: Why we need both request error and service error?

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -0,0 +1,95 @@
...
+        mapping_name: str = None,
+    ):
+        stream_format = stream_format.value if isinstance(stream_format, DataFormat) else DataFormat(stream_format.lower()).value
+        endpoint = self._streaming_ingest_endpoint + database + "/" + table + "?streamFormat=" + stream_format
+        if mapping_name is not None:
+            endpoint = endpoint + "&mappingName=" + mapping_name
+
+        await self._execute(endpoint, database, None, stream, self._streaming_ingest_default_timeout, properties)
+
+    async def _query(self, endpoint: str, **kwargs) -> list:
```
ArielYehezkely, line 65: this is confusing, why split _execute to three methods? (_execute, _query, _run_request)

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -163,10 +192,11 @@ class MsiTokenProvider(TokenProviderBase):
     The args parameter is a dictionary conforming with the ManagedIdentityCredential initializer API arguments
     """
 
-    def __init__(self, kusto_uri: str, msi_args):
-        super().__init__(kusto_uri)
+    def __init__(self, is_async: bool, kusto_uri: str, msi_args):
```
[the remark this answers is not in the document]
AsafMah: Only ManagedIdentityCredential  has proper async support.
If msal would add it one day we could add it to the other ones, right now they default to the generic sync_to_async

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -58,6 +68,18 @@ def get_token(self):
 
         return self._valid_token_or_throw(token)
 
+    async def get_token_async(self):
+        """ Get a token asynchronously silently from cache or authenticate if cached token is not found """
+        if not self._initialized:
```
yogilad, line 73: Does this function require a lock now ?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -179,19 +209,29 @@ def context(self) -> dict:
 
     def _init_impl(self):
         try:
-            self.msi_auth_context = ManagedIdentityCredential(**self._msi_args)
+            if self.is_async:
```
yogilad, line 212: This is odd.
Suggest you lazy init the correct instance upon use instead of keeping this member

yogilad (commented):
Part #1 
I'll keep reading this on Sunday

AsafMah:
@yogilad  You're right, get_token should always have been protected with I lock.

The problem with lazy-initing get_token_cache for msi is that I'll need to lock it too.

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: @yogilad You're right, get_token should always have been protected with I lock.

The problem with lazy-initing get_token_cache for msi is that I'll need to lock it too.

.github/workflows/pythonpackage.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8]
```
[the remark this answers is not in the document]
AsafMah: This is the minimum version the library supports.
3.5 is in EOL, while it is breaking, we really need to consider this.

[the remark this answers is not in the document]
ArielYehezkely: Maybe check python version and except if user tries to use the async client, like the optional packages.

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Sorry - I added the lock to get_token but not to get_token_async, I thought I did and that's why I resolved it

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -39,22 +50,38 @@ class TokenProviderBase(abc.ABC):
...
+        self.is_async = is_async
         self._kusto_uri = kusto_uri
         if kusto_uri is not None:
             self._scopes = [kusto_uri + ".defult" if kusto_uri.endswith("/") else kusto_uri + "/.default"]
             self._cloud_info = CloudSettings.get_cloud_info()
 
     def get_token(self):
         """ Get a token silently from cache or authenticate if cached token is not found """
+
+        with TokenProviderBase.lock:
```
yogilad, line 65: Pleas use a double check lock pattern

AsafMah: Like this?

yogilad: Yes

yihezkel: The double-checked locking pattern is a useful pattern in some languages, but notoriously not thread-safe in others.

Coincidentally, our own Michael Spector suggested in https://stackoverflow.com/questions/55957708/is-double-checked-locking-thread-safe-in-python that the value of this pattern depends on the specific Python implementation.

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -178,20 +217,39 @@ def context(self) -> dict:
...
     def _init_impl(self):
-        try:
-            self.msi_auth_context = ManagedIdentityCredential(**self._msi_args)
-        except Exception as e:
-            raise KustoClientError("Failed to initialize MSI ManagedIdentityCredential with [" + str(self._msi_params) + "]\n" + str(e))
+        pass
 
     def _get_token_impl(self) -> dict:
-        return None
+        try:
```
ArielYehezkely, line 223: Catch by exception type or add a variable to decide which error message to throw.

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -726,67 +782,25 @@ def execute_streaming_ingest(
...
-
+        response_json = response.json()
         try:
-            raise KustoServiceError([response.json()], response)
-        except ValueError:
-            if response.text:
-                raise KustoServiceError([response.text], response)
-            else:
-                raise KustoServiceError("Server error response contains no data.", response)
+            response.raise_for_status()
```
ArielYehezkely, line 802: Consistency - in aio client you check `if response.status != 200:`

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -0,0 +1,87 @@
...
+from ..client import KustoClient as KustoClientSync, _KustoClientBase, KustoConnectionStringBuilder, ClientRequestProperties, ExecuteRequestParams
+from ..response import KustoResponseDataSet
+from ..security import _AadHelper
+
+try:
+    from aiohttp import ClientResponse, ClientSession
+except ImportError:
+    raise KustoAioSyntaxError()
+
+ResponseTuple = namedtuple("ResponseTuple", ["response", "response_json"])
```
ArielYehezkely, line 18: Do you use it elsewhere?

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -0,0 +1,87 @@
...
+        timeout = request_params.timeout
+
+        if self._auth_provider:
+            request_headers["Authorization"] = await self._auth_provider.acquire_authorization_header_async()
+
+        async with self._session.post(endpoint, headers=request_headers, data=payload, json=json_payload, timeout=timeout.seconds) as response:
+            response_json = await response.json()
+
+            if response.status != 200:
+                raise KustoServiceError([response_json], response)
```
ArielYehezkely, line 85: In the sync client you create it as follows `KustoServiceError(response_json, response)`

ArielYehezkely (changes requested):

azure-kusto-data/azure/kusto/data/_token_providers.py
ArielYehezkely: Sorry, I mean, instead of the try inside try..

AsafMah: got it, should  be good now

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: not resolved..

AsafMah: Sorry, I misunderstood it before, I thought this was about the async client not matching the client, but now I see that I changed the normal client's behavior too..
I'll fix it

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -571,7 +614,57 @@ def init_poolmanager(self, *args, **kwargs):
...
+            if payload:
+                raise KustoServiceError("The ingestion endpoint does not exist. Please enable streaming ingestion on your cluster.", response) from exception
+            else:
+                raise KustoServiceError("The requested endpoint '{}' does not exist.".format(endpoint), response) from exception
+        if payload:
+            raise KustoServiceError(
+                "An error occurred while trying to ingest: Status: {0.status_code}, Reason: {0.reason}, Text: {1}.".format(response, response_text), response
+            ) from exception
+        try:
+            raise KustoServiceError([response_json], response) from exception
```
ArielYehezkely, line 659: That is super weird.. let talk offline..

AsafMah: I think it's good now

ArielYehezkely (dismissed):

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -558,6 +563,43 @@ def to_json(self):
         return json.dumps({"Options": self._options, "Parameters": self._parameters}, default=str)
 
 
+class ExecuteRequestParams:
```
danield137, line 566: I'm not sure why you need a class here

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -0,0 +1,86 @@
...
+
+        await self._execute(endpoint, database, None, stream, self._streaming_ingest_default_timeout, properties)
+
+    @aio_documented_by(KustoClientSync._execute)
+    async def _execute(
+        self, endpoint: str, database: str, query: Optional[str], payload: Optional[io.IOBase], timeout: timedelta, properties: ClientRequestProperties = None
+    ) -> KustoResponseDataSet:
+
+        request_params = ExecuteRequestParams(database, payload, properties, query, timeout, self._request_headers)
+        json_payload = request_params.json_payload
```
danield137, line 71: not sure why you expand these here instead of passing `**request_params` below

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -0,0 +1,86 @@
...
+        if self._auth_provider:
+            request_headers["Authorization"] = await self._auth_provider.acquire_authorization_header_async()
+
+        async with self._session.post(endpoint, headers=request_headers, data=payload, json=json_payload, timeout=timeout.seconds) as response:
+            response_json = None
+            try:
+                response_json = await response.json()
+                response.raise_for_status()
+            except Exception as e:
+                self._handle_http_error(e, endpoint, payload, response, response_json, await response.text())
```
danield137, line 84: can't ` await response.text()` raise?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -558,6 +563,43 @@ def to_json(self):
...
+            request_headers["Content-Encoding"] = "gzip"
+        request_headers["x-ms-client-request-id"] = client_request_id_prefix + str(uuid.uuid4())
+        if properties is not None:
+            if properties.client_request_id is not None:
+                request_headers["x-ms-client-request-id"] = properties.client_request_id
+            if properties.application is not None:
+                request_headers["x-ms-app"] = properties.application
+            if properties.user is not None:
+                request_headers["x-ms-user"] = properties.user
+            if properties.get_option(ClientRequestProperties.no_request_timeout_option_name, False):
```
danield137, line 591: https://github.com/Azure/azure-kusto-python/blob/a6e41842810ab2b82e81a826dd91fd5f9a6e70a2/azure-kusto-data/azure/kusto/data/client.py#L535

danield137 (dismissed):
Looks good, minor comments

azure-kusto-data/azure/kusto/data/client.py
AsafMah: This logic was here before the PR, but I assume that this option can be set, but set to False, so has_option isn't enough

azure-kusto-data/azure/kusto/data/client.py
AsafMah: What do you think would be good instead?

Have a function that returns a 3 tuple?
Or split it into multiple functions that you have to call in the correct order (since timeout is dependent on the headers)?

I do agree it's a bit ugly, but I don't have a better idea..

azure-kusto-data/azure/kusto/data/aio/client.py
AsafMah: request_params isn't a dictionary so I can't splat it.
Or do you mean just removing the interim variables?

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/aio/client.py
danield137: I think I would just avoid the class and destructre it

woanware:
This appears to have the same bug as the previous PR?

To fix the "too many file selectors" error I had to change the following line in "class _KustoClientBase", since it kept the HTTP socket open:

```
```

To:

```
self._request_headers = {'Connection':'close', "Accept": "application/json", "Accept-Encoding": "gzip,deflate", "x-ms-client-version": "Kusto.Python.Client:" + VERSION}
```

danield137 (dismissed):

AsafMah:
@woanware 
Setting the header is a band-aid solution, and we generally don't want to do that, since it means you have to open a new connection for every request and it is also removed in HTTP/2.

If we are leaking connections here that's a serious problem, I'll try checking myself but if you have minimal reproducing code that would help a lot.

The previous implementation also opened a new session for each connection, which could also have been a cause for that problem. 
This one only opens a session per Client, so see if this change makes a difference for you.

danield137 (dismissed):

ArielYehezkely (approved):

PR #298, opened 2021-01-19
Cleanup types

AsafMah:
#### Pull Request Description

Went through the repo with Pycharm,  and I found some small errors, mostly related to Types
I fixed them in this small PR

- Added proper imports and names whenever I can
- Specified Optional when needed
- Fixed typos in members
- Updated deprecated function

---

#### Future Release Comment
**Fixes:**
- Added proper imports and names whenever I can
- Specified Optional when needed
- Fixed typos in members
- Updated deprecated function

PR #301, opened 2021-02-11
Fix support for unidentifiable OSes

yihezkel:
#### Pull Request Description

When setting TCP Keep-Alive socket options, return empty list if we can't identify the OS.

---

#### Future Release Comment
**Fixes:**
- When setting TCP Keep-Alive socket options, return empty list if we can't identify the OS.


ArielYehezkely (approved):
Did you see Asaf's PR #298?

PR #303, opened 2021-02-25
MSAL Interactive Login

yogilad:
#### Pull Request Description

Add an option to login interactively with MSAL

---

#### Future Release Comment
**Breaking Changes:**
- Add an option to login interactively with MSAL (MSAL version changed to 1.9.0 or higher)
`KustoConnectionStringBuilder.with_interactive_login(service_url, authority, optional_login_hint, optional_domain_hint)`

**Bug Fixes:** 
* Exception is thrown when reporting exceptions during MSI authentication

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -370,6 +370,43 @@ class DeviceLoginTokenProvider(TokenProviderBase):
         return self._valid_token_or_none(token)
 
 
+class InteractiveLoginTokenProvider(TokenProviderBase):
+    """ Acquire a token from MSAL with Device Login flow """
+
+    def __init__(self, kusto_uri: str, authority_uri: str, login_hint: str = None, domain_hint: str = None):
+        super().__init__(kusto_uri)
+        self._msal_client = None
+        self._auth = authority_uri
+        self._login_hint = login_hint
+        self._domain_hint = domain_hint
+        self._account = None
+
+    @staticmethod
+    def name() -> str:
+        return "DeviceLoginTokenProvider"
+
+    def context(self) -> dict:
+        return {"authority": self._auth, "client_id": self._cloud_info.kusto_client_app_id}
+
+    def _init_impl(self):
+        self._msal_client = PublicClientApplication(client_id=self._cloud_info.kusto_client_app_id, authority=self._auth)
+
+    def _get_token_impl(self) -> dict:
+        token = self._msal_client.acquire_token_interactive(scopes=self._scopes, prompt="select_account", login_hint=self._login_hint,
+                                                            domain_hint=self._domain_hint)
+        return self._valid_token_or_throw(token)
+
+    def _get_token_from_cache_impl(self) -> dict:
+        account = None
+        if self._login_hint is not None:
+            accounts = self._msal_client.get_accounts(self._login_hint)
+            if len(accounts) > 0:
+                account = accounts[0]
+
+        token = self._msal_client.acquire_token_silent(scopes=self._scopes, account=account)
+        return self._valid_token_or_none(token)
+
+
 class ApplicationKeyTokenProvider(TokenProviderBase):
     """ Acquire a token from MSAL with application Id and Key """
 
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -383,7 +383,7 @@ class InteractiveLoginTokenProvider(TokenProviderBase):
 
     @staticmethod
     def name() -> str:
-        return "DeviceLoginTokenProvider"
+        return "InteractiveLoginTokenProvider"
 
     def context(self) -> dict:
         return {"authority": self._auth, "client_id": self._cloud_info.kusto_client_app_id}
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License
 import abc
 import webbrowser
-from typing import Callable
+from typing import Callable, Optional
 from msal import ConfidentialClientApplication, PublicClientApplication
 from azure.identity import ManagedIdentityCredential
 from .exceptions import KustoClientError
@@ -21,6 +21,7 @@ class TokenConstants:
     MSAL_PUBLIC_CERT = "public_certificate"
     MSAL_DEVICE_MSG = "message"
     MSAL_DEVICE_URI = "verification_uri"
+    MSAL_INTERACTIVE_PROMPT = "select_account"
     AZ_TOKEN_TYPE = "tokenType"
     AZ_ACCESS_TOKEN = "accessToken"
     AZ_REFRESH_TOKEN = "refreshToken"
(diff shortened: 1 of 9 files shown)


azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -370,6 +370,43 @@ def _get_token_from_cache_impl(self) -> dict:
...
+        super().__init__(kusto_uri)
+        self._msal_client = None
+        self._auth = authority_uri
+        self._login_hint = login_hint
+        self._domain_hint = domain_hint
+        self._account = None
+
+    @staticmethod
+    def name() -> str:
+        return "DeviceLoginTokenProvider"
```
ArielYehezkely, line 386: InteractiveLoginTokenProvider

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -370,6 +370,43 @@ def _get_token_from_cache_impl(self) -> dict:
...
+        )
+        return self._valid_token_or_throw(token)
+
+    def _get_token_from_cache_impl(self) -> dict:
+        account = None
+        accounts = self._msal_client.get_accounts(self._login_hint)
+        if len(accounts) > 0:
+            account = accounts[0]
+
+        token = self._msal_client.acquire_token_silent(scopes=self._scopes, account=account)
```
ArielYehezkely, line 406: Are you sure that acquire_token_silent doesn't throw when account is None?
Maybe worth checking before and return None

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: On the first pass before the prompt is called there are no accounts.
Token acquisition always starts with the cache and then goes to the actual API, so this happens every time.

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -370,6 +370,43 @@ def _get_token_from_cache_impl(self) -> dict:
         return self._valid_token_or_none(token)
 
 
+class InteractiveLoginTokenProvider(TokenProviderBase):
+    """ Acquire a token from MSAL with Device Login flow """
+
+    def __init__(self, kusto_uri: str, authority_uri: str, login_hint: str = None, domain_hint: str = None):
```
AsafMah, line 376: Use Optional[str] for parameters that can be None

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -370,6 +370,43 @@ def _get_token_from_cache_impl(self) -> dict:
         return self._valid_token_or_none(token)
 
 
+class InteractiveLoginTokenProvider(TokenProviderBase):
+    """ Acquire a token from MSAL with Device Login flow """
+
+    def __init__(self, kusto_uri: str, authority_uri: str, login_hint: str = None, domain_hint: str = None):
+        super().__init__(kusto_uri)
+        self._msal_client = None
```
AsafMah, line 378: consider adding type hints to the class members (see my type hinting PR, maybe we want to wait for that to merge)

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -370,6 +370,43 @@ def _get_token_from_cache_impl(self) -> dict:
...
+
+    def context(self) -> dict:
+        return {"authority": self._auth, "client_id": self._cloud_info.kusto_client_app_id}
+
+    def _init_impl(self):
+        self._msal_client = PublicClientApplication(client_id=self._cloud_info.kusto_client_app_id, authority=self._auth)
+
+    def _get_token_impl(self) -> dict:
+        token = self._msal_client.acquire_token_interactive(
+            scopes=self._scopes, prompt="select_account", login_hint=self._login_hint, domain_hint=self._domain_hint
```
AsafMah, line 396: "select_account" - should be a const

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -492,6 +512,19 @@ def az_cli(self):
     def token_provider(self):
         return self._token_provider
 
+    @property
```
AsafMah, line 515: consider annotating return types

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -492,6 +512,19 @@ def az_cli(self):
     def token_provider(self):
         return self._token_provider
 
+    @property
+    def interactive_login(self):
+        val = self._internal_dict.get(self.ValidKeywords.interactive_login)
+        return val is not None and val
```
AsafMah, line 518: I'm pretty sure that `bool(val)` would achive the same effect and be more readable

test.bat
```diff
@@ -1,7 +1,8 @@
 cd %PROJECTS_HOME%\azure-kusto-python
 call workon kusto
 call pip uninstall azure-kusto-data azure-kusto-ingest -y 
-call pip install ./azure-kusto-data[pandas] ./azure-kusto-ingest[pandas] 
+call pip install -e ./azure-kusto-data[pandas]
+call pip install -e ./azure-kusto-ingest[pandas]
```
AsafMah, line 5: Is there a reason to separate them?

azure-kusto-data/azure/kusto/data/exceptions.py
```diff
@@ -49,8 +49,8 @@ def __init__(self, authentication_method: str, exception: Exception, **kwargs):
         self.exception = exception
         if "authority" in kwargs:
             self.authority = kwargs["authority"]
-        if "resource" in kwargs:
-            self.kusto_cluster = kwargs["resource"]
+        if "kusto_uri" in kwargs:
+            self.kusto_cluster = kwargs["kusto_uri"]
```
AsafMah, line 53: This is technically a breaking change, no? Someone who inspects the exception will suddenly get none.
Maybe we can keep both for now and remove "resource" in the next major.

azure-kusto-data/azure/kusto/data/client.py
yogilad: out of the scope of this PR.
Plus it will only conflict with your changes more if I annotate APIs which you've annotated in your PR

azure-kusto-data/azure/kusto/data/exceptions.py
yogilad: I got a comment from a customer this is not an intuitive name if the value is missing or corrupt (which was the case with his problem).
@ArielYehezkely , how do you feel about this change?

test.bat
yogilad: Yes, for some reason the -e flag does not work well with multiple resources

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: Syntax supported in 3.5?

AsafMah: With 3.5 you'll have to do the comment syntax, if you don't like it we can wait to 3.6

azure-kusto-data/azure/kusto/data/client.py
AsafMah: You resolved this and previous comment, but I don't see any comment from you or change in the code

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: Will wait

azure-kusto-data/azure/kusto/data/exceptions.py
ArielYehezkely: I don't think we need to worry about it, its not an API and I don't think people actually base their code on such members, people usually just print/log the exception.
If you want to be extra cautious you can add instead of replace but please open a PR which removes the old one with "merge with next major version bump" label.

AsafMah (approved):

PR #311, opened 2021-03-25
Fixed import on non-async machines

AsafMah:
Per the email, we can't even import the async managed identity if the aiohttp package isn't installed.
This fixes that like the other method - with a shim that throws when called, which in sync code never will.


azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -21,6 +20,15 @@ def sync_to_async(f):
         raise KustoAioSyntaxError()
 
 
+try:
+    from azure.identity.aio import ManagedIdentityCredential as AsyncManagedIdentityCredential
+except ImportError:
+
+    class AsyncManagedIdentityCredential:
+        def __init__(self):
+            raise KustoAioSyntaxError()
```
yogilad, line 29: No message or anything?

PR #315, opened 2021-05-03
Add streaming query

AsafMah:
#### Pull Request Description

Adding streaming query api, for both async and non-async clients.
---

#### Future Release Comment

**Features:**
- Adding streaming query api, for both async and non-async clients.


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -853,19 +853,61 @@ def execute_streaming_ingest(
 
         self._execute(endpoint, database, None, stream, self._streaming_ingest_default_timeout, properties)
 
-    def _execute(
-        self, endpoint: str, database: str, query: Optional[str], payload: Optional[io.IOBase], timeout: timedelta, properties: ClientRequestProperties = None
+    def execute_streaming_query(
```
ArielYehezkely, line 856: Can't we just pass a boolean argument of 'streaming'?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -853,19 +853,61 @@ def execute_streaming_ingest(
...
+        response = self._retrieve_response(self._query_endpoint, database, properties, query, None, timeout, stream=True)
+
+        try:
+            response.raise_for_status()
+        except Exception as e:
+            self._handle_http_error(e, self._query_endpoint, None, response, response.json(), response.text)
+
+        return response
+
+    def _retrieve_response(
```
ArielYehezkely, line 881: Can't it be part of _execute?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -853,19 +853,61 @@ def execute_streaming_ingest(
 
         self._execute(endpoint, database, None, stream, self._streaming_ingest_default_timeout, properties)
 
-    def _execute(
-        self, endpoint: str, database: str, query: Optional[str], payload: Optional[io.IOBase], timeout: timedelta, properties: ClientRequestProperties = None
+    def execute_streaming_query(
+        self, database: str, query: str, timeout: timedelta = _KustoClientBase._query_default_timeout, properties: Optional[ClientRequestProperties] = None
+    ) -> Response:
```
ArielYehezkely, line 858: I don't understand the result - in the non-streaming query we return a wrapper of type "KustoResponseDataSet" to make it easy to use the response as tables and provide type conversions.
I think the raw result will be hard to use unless I miss something here.

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -853,19 +853,61 @@ def execute_streaming_ingest(
...
+            self._handle_http_error(e, self._query_endpoint, None, response, response.json(), response.text)
+
+        return response
+
+    def _retrieve_response(
+        self,
+        endpoint: str,
+        database: str,
+        properties: Optional[ClientRequestProperties],
+        query: str,
```
ArielYehezkely, line 886: Usually we place query before ClientRequestProperties

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -853,19 +853,61 @@ def execute_streaming_ingest(
...
+        payload: Optional[io.IOBase],
+        timeout: timedelta,
+        stream: bool,
```
ArielYehezkely, line 889: Set a default

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -853,19 +853,61 @@ def execute_streaming_ingest(
...
+        return response
 
-        response = self._session.post(endpoint, headers=request_headers, data=payload, json=json_payload, timeout=timeout.seconds)
+    def _execute(
+        self,
+        endpoint: str,
+        database: str,
+        query: Optional[str],
+        payload: Optional[io.IOBase],
+        timeout: timedelta,
```
ArielYehezkely, line 906: I believe timeout is also an optional

azure-kusto-ingest/tests/e2e.py
```diff
@@ -218,6 +224,28 @@ async def assert_rows_added(cls, expected: int, timeout=60):
         cls.current_count += actual
         assert actual == expected, "Row count expected = {0}, while actual row count = {1}".format(expected, actual)
 
+    def test_streaming_query(self):
+        with pytest.raises(KustoServiceError):
+            self.client.execute_query(self.test_db, self.streaming_test_table_query)
```
ArielYehezkely, line 229: Can you please add a comment why it should fail?

azure-kusto-data/azure/kusto/data/client.py
AsafMah: The return type of _execute is different than that of execute_streaming_query,
I think it's a bad form to change the return type of a function based on a parameter

azure-kusto-data/azure/kusto/data/client.py
AsafMah: It's for sharing code between _execute and execute_streaming_query

azure-kusto-data/azure/kusto/data/client.py
AsafMah: This is modeled after the java api - https://github.com/Azure/azure-kusto-java/pull/170

To convert it to KustoResponseDataSet we need all of the data in-memory, which of course isn't possible for a huge amount of data.
The usecase is for them to dump it into a file and parse the json there.

@yihezkel correct me if I'm wrong

azure-kusto-data/azure/kusto/data/client.py
AsafMah: Why?

It's a private function, so it won't help the user to set it, and for us it makes sure we don't accidentally miss the parameter

azure-kusto-data/azure/kusto/data/client.py
AsafMah: See before

ArielYehezkely: Sorry -  
Not yours: I believe timeout is also an optional

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: _execute was the method for sharing code between all the "public" methods - so now its not good enough so we add _execute2? - I don't like it

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: It comes from the same place as the previous comment - we should have 1 _execute with default as most of the flows use stream=false

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: Same discussion as the result type

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: I don't think its useful. 
I did some work to parse the result in lazy manner, maybe we can find the gap between what we have today and full lazy evaluation and make everything "streaming"

yihezkel: The Java SDK was a special case. As discussed, I recommend using streaming object creation to create a KustoResponseDataSet incrementally. I'm not familiar with Python support for this concept, but hopefully we wou;dn't need to change the datatype to support this.

ArielYehezkely (commented):
Overall looks good!
Few general points:

1. There are many parts and new classes, I think we need some UT to be confident that each element does what's expected from it. (for example JsonTokenReader.)
2. I see new exceptions, please make sure they are used to surface a real issue to the user and not just used to communicate between internal parts.
3. Need to think about aggregating exceptions, so if an error occurred but we can keep reading, it won't stop the iteration. (Can be a later improvement, not part of this PR)

azure-kusto-data/azure/kusto/data/_models.py
```diff
@@ -151,3 +156,62 @@ def __bool__(self):
...
+    @property
+    def columns_count(self) -> int:
+        return len(self.columns)
+
+    def to_dict(self):
+        """Converts the table to a dict."""
+        return {"name": self.table_name, "kind": self.table_kind}
+
+    def __len__(self):
+        if not self.finished:
```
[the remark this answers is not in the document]
AsafMah: Not sure what you mean, in KustoResultTable it's:
```python
    def __len__(self):
        return self.rows_count
```

azure-kusto-data/azure/kusto/data/aio/streaming_response.py
```diff
@@ -0,0 +1,253 @@
...
+
+    async def skip_to_end(self):
+        async for _ in self.json_iter:
+            pass
+
+    async def skip_children(self, prev_token: JsonToken):
+        if prev_token.token_type == JsonTokenType.MAP_KEY:
+            await self.read_next_token_or_throw()
+        elif prev_token.token_type in JsonTokenType.start_tokens():
+            while True:
```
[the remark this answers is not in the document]
AsafMah: Python doesn't have a do-while statement, do you mean to just restructure it so it won't be while true?

azure-kusto-data/azure/kusto/data/aio/streaming_response.py
```diff
@@ -0,0 +1,253 @@
...
+            return res
+
+    async def row_iterator(self):
+        await self.reader.read_token_of_type(JsonTokenType.START_ARRAY)
+        while True:
+            token = await self.reader.read_token_of_type(JsonTokenType.START_ARRAY, JsonTokenType.END_ARRAY, JsonTokenType.START_MAP)
+            if token.token_type == JsonTokenType.START_MAP:
+                raise KustoServiceError(
+                    "Received error in data: " + str(self.parse_object(skip_start=True))
+                )  # Todo - replace this with a better error once managed streaming ingest is merged
```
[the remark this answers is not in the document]
AsafMah: I'm doing a managed streaming ingest branch, and in it I created more specific errors so it can parse them for its functions, so I added this comment to not forget adding them here

This  is probably premature since that branch's PR isn't up yet

azure-kusto-data/azure/kusto/data/aio/streaming_response.py
```diff
@@ -0,0 +1,253 @@
...
+        if prev_token.token_type == JsonTokenType.MAP_KEY:
+            await self.read_next_token_or_throw()
+        elif prev_token.token_type in JsonTokenType.start_tokens():
+            while True:
+                potential_end_token = await self.read_next_token_or_throw()
+                if potential_end_token.token_path == prev_token.token_path and potential_end_token.token_type in JsonTokenType.end_tokens():
+                    break
+
+    async def skip_until_property_name(self, name: str):
+        while True:
```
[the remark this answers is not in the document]
AsafMah: I replaced a lot of `while True` loops with for loops (like in `skip_until_property_name_or_end_object`) when possible,
but here I'm not sure what's the better way to do it, other than duplicating `skip_until_property_name_or_end_object`.
Or maybe that's ok?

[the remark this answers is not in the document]
ArielYehezkely: That's ok

azure-kusto-data/azure/kusto/data/aio/streaming_response.py
```diff
@@ -0,0 +1,248 @@
...
+            raise Exception(f"Unexpected token {token}")
+
+    async def skip_until_token_with_paths(self, *tokens: (JsonTokenType, str)):
+        async for token in self:
+            if any((token.token_type == t_type and token.token_path == t_path) for (t_type, t_path) in tokens):
+                return token
+            await self.skip_children(token)
+
+
+class ProgressiveDataSetEnumerator:
```
ArielYehezkely, line 100: Amit Ofir brought to my knowledge that streaming isn't progressive - to have a progressive query you need to specify it in the ClientRequestProperties  - but this isn't the case here, so I think we better not use the term "progressive" as this is confusing anyway. Sorry if I were misleading you..

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/exceptions.py
```diff
@@ -16,6 +16,16 @@ class KustoError(Exception):
     """Base class for all exceptions raised by the Kusto Python Client Libraries."""
 
 
+class KustoStreamingQueryError(KustoError):
+    def __init__(self, *args: object) -> None:
+        super().__init__(*args)
+
+
+class KustoTokenParsingError(KustoStreamingQueryError):
+    def __init__(self, *args: object) -> None:
```
danield137, line 25: hmm, what about `...`? doesn't it automatically call the super?

azure-kusto-data/azure/kusto/data/response.py
```diff
@@ -90,6 +74,30 @@ def __len__(self) -> int:
...
+        self.tables = [KustoResultTable(t) for t in json_response]
+        self.tables_count = len(self.tables)
+        self.tables_names = [t.table_name for t in self.tables]
+
+    @property
+    def primary_results(self) -> List[KustoResultTable]:
+        """Returns primary results. If there is more than one returns a list."""
+        if self.tables_count == 1:
+            return self.tables
+        primary = list(filter(lambda x: x.table_kind == WellKnownDataSet.PrimaryResult, self.tables))
```
danield137, line 96: why not a list comprehension instead of the list/filter?

azure-kusto-data/azure/kusto/data/response.py
```diff
@@ -138,3 +146,88 @@ class KustoResponseDataSetV2(KustoResponseDataSet):
 
     def __init__(self, json_response: List[dict]):
         super(KustoResponseDataSetV2, self).__init__([t for t in json_response if t["FrameType"] == "DataTable"])
+
+
+class KustoStreamingResponseDataSet(BaseKustoResponseDataSet):
```
danield137, line 151: `Progressive`?

azure-kusto-data/azure/kusto/data/response.py
AsafMah: image ;)

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -879,19 +880,53 @@ def execute_streaming_ingest(
...
-
-        response = self._session.post(endpoint, headers=request_headers, data=payload, json=json_payload, timeout=timeout.seconds)
+        response = self._session.post(endpoint, headers=request_headers, json=json_payload, data=payload, timeout=timeout.seconds, stream=stream_response)
+
+        if stream_response:
+            try:
+                response.raise_for_status()
+                return response
+            except Exception as e:
+                self._handle_http_error(e, self._query_endpoint, None, response, response.json(), response.text)
```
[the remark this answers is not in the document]
AsafMah: self._handle_http_error always raises, so I'm not sure what the right thing to do here.
If we get to the next line it's a bug, so maybe doing an `assert false` or raising some kind of critical exception?

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -879,19 +880,53 @@ def execute_streaming_ingest(
...
         request_headers = request_params.request_headers
         timeout = request_params.timeout
-
         if self._auth_provider:
             request_headers["Authorization"] = self._auth_provider.acquire_authorization_header()
-
-        response = self._session.post(endpoint, headers=request_headers, data=payload, json=json_payload, timeout=timeout.seconds)
+        response = self._session.post(endpoint, headers=request_headers, json=json_payload, data=payload, timeout=timeout.seconds, stream=stream_response)
+
+        if stream_response:
```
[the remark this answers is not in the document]
AsafMah: Not sure I understand.

Not having the whole request on memory was what prompted the PR, so it is a priority.

But we do read the json itself iteratively here byte by byte.
You block when reading a chunk, but something somewhere has to block

Did you mean something like a thread to read in the background and push updates?

azure-kusto-data/azure/kusto/data/aio/streaming_response.py
AsafMah: I removed the progressive wording, supporting it seems nice but I think it's Out Of Scope for this PR

@yogilad what do you think?

azure-kusto-data/azure/kusto/data/response.py
AsafMah: I removed the progressive wording, see above

danield137 (approved):

danield137 (approved):

PR #321, opened 2021-05-21
Return awaitable object from aio client __aexit__

artunduman:
#### Bug Fix

\_\_aexit\_\_ should return an awaitable.


AsafMah (approved):
Great catch!

PR #324, opened 2021-06-01
Fixed wrong kwargs and bumped version for release

AsafMah:
Fixing the typo in azcli, and gearing up to release 2.1.2 - there are clients waiting for the fix


yogilad (approved):
Correct, I fixed this in the sync command, but not in the async one

Issue #314, opened 2021-04-26
Add support for usgovernment regions kusto 


amankohli:
We need support on the module for usgov regions like usgovvirginia,usgovtexas etc

vladikbr:
You actually mean support for sovereign/air-gapped clouds.
We are working on this, as a workaround, you can obtain the token for your Kusto resource yourself and pass it to the with_token_provider method.

amankohli:
Yes I mean support for sovereign/air-gapped clouds. Do we have a timeline when the support will be added to the library?

yogilad:
Versions following change #325 should handle other clouds well.
If there are issues, let us know.

PR #325, opened 2021-06-08
Add dynamic metadata fetching

AsafMah:
#### Pull Request Description

_[Add a description of your pull request here]_

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- #314

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_cloud_settings.py
+++ b/azure-kusto-data/azure/kusto/data/_cloud_settings.py
@@ -18,8 +18,15 @@ FIRST_PARTY_AUTHORITY_URL = "https://login.microsoftonline.com/f8cdef31-a31e-4b4
 class CloudInfo:
     """This class holds the data for a specific cloud instance."""
 
-    def __init__(self, login_endpoint: str, login_mfa_required: bool, kusto_client_app_id: str, kusto_client_redirect_uri: str, kusto_service_resource_id: str,
-                 first_party_authority_url: str):
+    def __init__(
+        self,
+        login_endpoint: str,
+        login_mfa_required: bool,
+        kusto_client_app_id: str,
+        kusto_client_redirect_uri: str,
+        kusto_service_resource_id: str,
+        first_party_authority_url: str,
+    ):
         self.login_endpoint = login_endpoint
         self.login_mfa_required = login_mfa_required
         self.kusto_client_app_id = kusto_client_app_id
@@ -44,7 +51,7 @@ class CloudSettings:
         kusto_client_app_id=KUSTO_CLIENT_APP_ID,
         kusto_client_redirect_uri=REDIRECT_URI,
         kusto_service_resource_id=KUSTO_SERVICE_RESOURCE_ID,
-        first_party_authority_url=FIRST_PARTY_AUTHORITY_URL
+        first_party_authority_url=FIRST_PARTY_AUTHORITY_URL,
     )
 
     @classmethod
--- a/azure-kusto-data/azure/kusto/data/_cloud_settings.py
+++ b/azure-kusto-data/azure/kusto/data/_cloud_settings.py
@@ -1,18 +1,20 @@
 import os
-from functools import lru_cache
 from typing import Optional
 from urllib.parse import urljoin
+from threading import Lock
 
 import requests
 
 from azure.kusto.data.exceptions import KustoServiceError
 
-AUTH_ENV_VAR_NAME = "AadAuthorityUri"
-KUSTO_CLIENT_APP_ID = "db662dc1-0cfe-4e1c-a843-19a68e65be58"
-PUBLIC_LOGIN_URL = "https://login.microsoftonline.com"
-REDIRECT_URI = "https://microsoft/kustoclient"
-KUSTO_SERVICE_RESOURCE_ID = "https://kusto.kusto.windows.net"
-FIRST_PARTY_AUTHORITY_URL = "https://login.microsoftonline.com/f8cdef31-a31e-4b4a-93e4-5f571e91255a"
+METADATA_ENDPOINT = "v1/rest/auth/metadata"
+
+DEFAULT_AUTH_ENV_VAR_NAME = "AadAuthorityUri"
+DEFAULT_KUSTO_CLIENT_APP_ID = "db662dc1-0cfe-4e1c-a843-19a68e65be58"
+DEFAULT_PUBLIC_LOGIN_URL = "https://login.microsoftonline.com"
+DEFAULT_REDIRECT_URI = "https://microsoft/kustoclient"
+DEFAULT_KUSTO_SERVICE_RESOURCE_ID = "https://kusto.kusto.windows.net"
+DEFAULT_FIRST_PARTY_AUTHORITY_URL = "https://login.microsoftonline.com/f8cdef31-a31e-4b4a-93e4-5f571e91255a"
 
 
 class CloudInfo:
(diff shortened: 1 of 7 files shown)


azure-kusto-data/azure/kusto/data/_cloud_settings.py
```diff
@@ -1,35 +1,79 @@
...
+from functools import lru_cache
+from typing import Optional
+from urllib.parse import urljoin
+
+import requests
+
+from azure.kusto.data.exceptions import KustoServiceError
 
 AUTH_ENV_VAR_NAME = "AadAuthorityUri"
 KUSTO_CLIENT_APP_ID = "db662dc1-0cfe-4e1c-a843-19a68e65be58"
```
yogilad, line 11: This and below need a DEFAULT_ prefix

azure-kusto-data/azure/kusto/data/_cloud_settings.py
```diff
@@ -1,35 +1,79 @@
...
+
+    def authority_uri(self, authority_id: Optional[str]):
+        return self.login_endpoint + "/" + (authority_id or "organizations")
 
 
 class CloudSettings:
     """This class holds data for all cloud instances, and returns the specific data instance by parsing the dns suffix from a URL"""
 
-    _cloud_info = None
+    METADATA_ENDPOINT = "v1/rest/auth/metadata"
```
yogilad, line 44: nit: group constants together? (unless there's a reason for  this being separated)

azure-kusto-data/azure/kusto/data/_cloud_settings.py
```diff
@@ -1,35 +1,79 @@
...
-        cls._init_once()
-        return cls._cloud_info
+    @lru_cache(maxsize=None)
+    def get_cloud_info_for_cluster(cls, kusto_uri: str) -> CloudInfo:
+        result = requests.get(urljoin(kusto_uri, cls.METADATA_ENDPOINT))
+
+        if result.status_code == 200:
+            content = result.json()
+            if content is None or content == {}:
+                raise KustoServiceError("Kusto returned an invalid cloud metadata response", result)
```
yogilad, line 65: Should we not fallback here as well?
How does C# handle this case?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -61,16 +61,23 @@ class TokenProviderBase(abc.ABC):
...
+
+            self._scopes = [resource_uri + "/.default"]
 
     def get_token(self):
         """Get a token silently from cache or authenticate if cached token is not found"""
 
         if not self._initialized:
             with TokenProviderBase.lock:
                 if not self._initialized:
+                    self.init_cloud()
```
yogilad, line 80: a) You need to be absolutely sure cloud info is not used in the initializers of serviced class
b) I think this is hard constraint, and in most cases it would be served from cache. Consider moving back to __init__

azure-kusto-data/azure/kusto/data/_cloud_settings.py
```diff
@@ -1,35 +1,79 @@
...
+        kusto_service_resource_id=KUSTO_SERVICE_RESOURCE_ID,
+        first_party_authority_url=FIRST_PARTY_AUTHORITY_URL,
+    )
 
     @classmethod
-    def get_cloud_info(cls) -> CloudInfo:
-        """Get the details of a cloud according to the DNS suffix of the provided connection string"""
-        cls._init_once()
-        return cls._cloud_info
+    @lru_cache(maxsize=None)
```
yogilad, line 58: How does this sorcery work

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -318,10 +326,10 @@ def name() -> str:
         return "UserPassTokenProvider"
 
     def context(self) -> dict:
-        return {"authority": self._auth, "client_id": self._cloud_info.kusto_client_app_id, "username": self._user}
+        return {"authority": self._cloud_info.authority_uri(self._auth), "client_id": self._cloud_info.kusto_client_app_id, "username": self._user}
```
yogilad, line 329: e.g. this could be an issue if context is called after a failure do to endpoint

azure-kusto-data/tests/test_security.py
```diff
@@ -1,12 +1,14 @@
...
-from azure.kusto.data.exceptions import KustoAuthenticationError
+
 from azure.kusto.data import KustoConnectionStringBuilder
-from azure.kusto.data.security import _AadHelper
 from azure.kusto.data._token_providers import *
+from azure.kusto.data.exceptions import KustoAuthenticationError
+from azure.kusto.data.security import _AadHelper
 
 KUSTO_TEST_URI = "https://thisclusterdoesnotexist.kusto.windows.net"
+KUSTO_TEST__WORKING_URI = "https://sdkse2etest.eastus.kusto.windows.net"
```
yogilad, line 11: nit: double underscore in name __

azure-kusto-data/azure/kusto/data/_cloud_settings.py
AsafMah: Yes, this behavior is taken directly from AuthenticationMetadataProvider.cs 
If we get 404 we return the default, but if we get 200 and a malformed answer we throw

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Putting it in init is wrong, because:
a. It can potentially take a long time, without the user explictly asking it 2
b. It would block on async flows, which is bad

So that needs to stay, but maybe there is something clever to do in regards to a) and the later context comment

azure-kusto-data/azure/kusto/data/_cloud_settings.py
AsafMah: It's a cool decorator from functools, it basically wraps the function and saves the result to a dict.

I've now decided to remove it, because it's impossible to control manually, which is needed for tests.
So now it's a simple dict+lock

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: You are right, flows like this could happen, and that's a big problem.

I see two options:
1. Add a call to init_cloud in the beginning of relevant context method.
The problem with that is the same as calling init_cloud in __init__ - it may unexpectedly block and we have no async equivalent
2.  Do a "best effort" - add a method to retrieve the cloud info, but only if it's already cached. Otherwise return None or the default cloud. This isn't great either since the users might get wrong/incomplete data.

I don't like either of these,  I'm not sure what's the correct way here.

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: I'd wrap this in a private init_once function and call it before Context.
You can have the base call it in Context() and make all derived classes implement context_impl()

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: (ATM) Context is called within auth requests to add, well... context to error exceptions.
I think (1) would be fine

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Is this ok?

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Is this ok?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -62,6 +63,32 @@ class TokenProviderBase(abc.ABC):
     def __init__(self, kusto_uri: str):
         self._kusto_uri = kusto_uri
 
+    def init_once(self, init_only_cloud=False):
```
yogilad, line 66: Should this be private?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -62,6 +63,32 @@ class TokenProviderBase(abc.ABC):
...
+                    self.init_cloud()
+                    self._cloud_initialized = True
+
+                if init_only_cloud:
+                    return
+
+                self._init_impl()
+                self._initialized = True
+
+    async def init_once_async(self):
```
yogilad, line 82: Should this be private?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -62,6 +63,32 @@ class TokenProviderBase(abc.ABC):
...
+        if not self._initialized:
+            with TokenProviderBase.lock:
+                if not self._initialized:
+                    if not self._cloud_initialized:
+                        await (sync_to_async(self.init_cloud)())
+                        self._cloud_initialized = True
+                    self._init_impl()
+                    self._initialized = True
+
     def init_cloud(self):
```
yogilad, line 92: Should this be private?
(I know it's mine :) )

yogilad (commented):
Seems ok with some minor comments

yogilad (dismissed):
Seems OK.
I would add explicit e2e tests for the endpoint

yogilad (approved):

PR #331, opened 2021-07-17
Fixed more lock errors

AsafMah:
This version moves the locks in token_providers to individual instances, rather than per class.

Since we don't use any shared class variables (as far as I can tell), this is safe to do and beneficial for performance.

For the async lock, this is essential - since the lock must be created with the same loop as the class, this means we can't initiate it until the class is created, and it requires different locks for instances inside different loops.

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -218,7 +218,7 @@ class BasicTokenProvider(TokenProviderBase):
 class CallbackTokenProvider(TokenProviderBase):
     """Callback Token Provider generates a token based on a callback function provided by the caller"""
 
-    def __init__(self, token_callback: Optional[Callable[[], str]], async_token_callback: Optional[Callable[[], Coroutine[None, None, str]]] = None):
+    def __init__(self, token_callback: Optional[Callable[[], str]], async_token_callback: Optional[Callable[[], Coroutine[None, None, str]]]):
         super().__init__(None)
         self._token_callback = token_callback
         self._async_token_callback = async_token_callback
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -60,20 +60,20 @@ class TokenProviderBase(abc.ABC):
     _cloud_info = None
     _scopes = None
 
-    # There are different locks for sync and async operations, since using a sync lock in an async context may cause a deadlock.
-    # This means that theoretically, if get_token() and get_token_async() were to be called at the same time, then there might be a race condition.
-    # Since this class is private, and the usage within the clients is limited to one type of function, this is ok to do.
-    lock = Lock()
-    async_lock = asyncio.Lock()
-
     def __init__(self, kusto_uri: str):
         self._kusto_uri = kusto_uri
+
+        # There are different locks for sync and async operations, since using a sync lock in an async context may cause a deadlock.
+        # This means that theoretically, if get_token() and get_token_async() were to be called at the same time, then there might be a race condition.
+        # Since this class is private, and the usage within the clients is limited to one type of function, this is ok to do.
+        self.async_lock = asyncio.Lock()
+        self.lock = Lock()
 
     def _init_once(self, init_only_cloud=False):
         if self._initialized:
             return
 
-        with TokenProviderBase.lock:
+        with self.lock:
             if self._initialized:
                 return
 
@@ -91,7 +91,7 @@ class TokenProviderBase(abc.ABC):
         if self._initialized:
             return
 
-        async with TokenProviderBase.async_lock:
+        async with self.async_lock:
             if self._initialized:
                 return
 
@@ -120,7 +120,7 @@ class TokenProviderBase(abc.ABC):
 
         token = self._get_token_from_cache_impl()
         if token is None:
-            with TokenProviderBase.lock:
+            with self.lock:
                 token = self._get_token_impl()
 
         return self._valid_token_or_throw(token)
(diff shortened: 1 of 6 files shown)


azure-kusto-data/tests/aio/test_async_token_providers.py
```diff
@@ -313,3 +313,25 @@ async def test_cloud_mfa_on(self):
...
+        This meant that if someone created a new event loop, and created a provider in it, awaiting on the lock would cause an exception because it belongs to
+        a different loop.
+        Now the lock is instantiated for every class instance, avoiding this issue.
+        """
+
+        async def start():
+            provider = MockProvider(KUSTO_URI)
+
+            async def inner():
+                async with provider.async_lock:
```
yogilad, line 330: Why is it public?

azure-kusto-data/tests/aio/test_async_token_providers.py
```diff
@@ -313,3 +313,25 @@ async def test_cloud_mfa_on(self):
...
+        """
+        This test makes sure that the lock inside of a TokenProvider, is created within the correct event loop.
+        Before this, the Lock was created once per class.
+        This meant that if someone created a new event loop, and created a provider in it, awaiting on the lock would cause an exception because it belongs to
+        a different loop.
+        Now the lock is instantiated for every class instance, avoiding this issue.
+        """
+
+        async def start():
+            provider = MockProvider(KUSTO_URI)
```
yogilad, line 327: So, noticed basic and callback providers are not async compatible.
a) why?
b) had callback been async you'd be able to implement this test by calling get token and having the async callback wait do async sleep

azure-kusto-data/tests/aio/test_async_token_providers.py
AsafMah: Good point, fixed now

azure-kusto-data/tests/aio/test_async_token_providers.py
AsafMah: All providers are async compatible, it's just that if they don't have their own implementation of _get_token_impl_async, it defaults to the base class one which just calls sync_to_async on it.
Since basic just returns a string it makes sense.

For callback while it technically works, it does seem like missing functionality.
So I added the option for an async callback, and changed the test like you suggested, please take a look

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -218,9 +218,10 @@ def _get_token_from_cache_impl(self) -> dict:
 class CallbackTokenProvider(TokenProviderBase):
     """Callback Token Provider generates a token based on a callback function provided by the caller"""
 
-    def __init__(self, token_callback: Callable[[], str]):
+    def __init__(self, token_callback: Optional[Callable[[], str]], async_token_callback: Optional[Callable[[], Coroutine[None, None, str]]] = None):
```
yogilad, line 221: This is an odd constructor.
I must provide a sync callable if I want to use the async one.
I can't think of a more reasonable solution. Perhaps __init__(self, tokenCallback, isAsync = false)

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -232,13 +233,23 @@ def _context_impl(self) -> dict:
...
+    @staticmethod
+    def _build_response(caller_token) -> dict:
         if not isinstance(caller_token, str):
             raise KustoClientError("Token provider returned something that is not a string [" + str(type(caller_token)) + "]")
 
         return {TokenConstants.MSAL_TOKEN_TYPE: TokenConstants.BEARER_TYPE, TokenConstants.MSAL_ACCESS_TOKEN: caller_token}
 
+    def _get_token_impl(self) -> dict:
+        if self._token_callback is None:
+            raise Exception("token_callback is None, can't retrieve token")
```
yogilad, line 245: KustoClientException

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: You can pass None for the sync callback

AsafMah: Either way, that's probably not the interface we should be worried about since it's not user-facing

`KustoConnectionStringBuilder.with_token_provider` is currently the same, but maybe we can split it to `with_token_provider` and `with_async_token_provider`?
It would prevent the user from supplying both, but I don't think that's a real usecase

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -386,18 +387,31 @@ def with_aad_managed_service_identity_authentication(
         return kcsb
 
     @classmethod
-    def with_token_provider(cls, connection_string: str, token_provider: Callable[[], str]) -> "KustoConnectionStringBuilder":
+    def with_token_provider(
+        cls,
+        connection_string: str,
+        token_provider: Optional[Callable[[], str]],
+        async_token_provider: Optional[Callable[[], Coroutine[None, None, str]]] = None,
+    ) -> "KustoConnectionStringBuilder":
```
yogilad, line 395: Same comment applies here.

AsafMah: see above

yogilad (changes requested):
Not a fan of the new signiture

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: makes more sense.
If you do init tokenProvider with Nones make both mandatory and use named arguments in the using code

AsafMah: Is this what you meant @yogilad ?

yogilad (approved):

PR #334, opened 2021-07-21
Loosen Python Pins for compatibility

dciborow:
#### Pull Request Description

Loosen Python Pins

---

#### Future Release Comment
Loosen Python pins.

### DIFF
--- a/azure-kusto-data/setup.py
+++ b/azure-kusto-data/setup.py
@@ -43,6 +43,6 @@ setup(
     namespace_packages=["azure"],
     keywords="kusto wrapper client library",
     packages=find_packages(exclude=["azure", "tests"]),
-    install_requires=["python-dateutil>=2.8.0", "requests>=2.13.0", "azure-identity~=1.5.0", "msal~=1.9.0"],
+    install_requires=["python-dateutil>=2.8.0", "requests>=2.13.0", "azure-identity>=1.5.0,<=1.6.0", "msal>=1.9.0,<2"],
     extras_require={"pandas": ["pandas"], "aio": ["aiohttp~=3.4.4", "asgiref~=3.2.3"]},
 )


dciborow:
#334

AsafMah:
Isn't  "msal~=1.9.0" equivalent to  "msal>=1.9.0,<2"?
And azure-identity stil feels restrictive, if they released a minor minor version of 1.6.1, we won't be able to use it

yogilad:
@AsafMah , see if we need to tweak this before permitting the merge

azure-kusto-data/setup.py
```diff
@@ -43,6 +43,6 @@
     namespace_packages=["azure"],
     keywords="kusto wrapper client library",
     packages=find_packages(exclude=["azure", "tests"]),
-    install_requires=["python-dateutil>=2.8.0", "requests>=2.13.0", "azure-identity~=1.5.0", "msal~=1.9.0"],
+    install_requires=["python-dateutil>=2.8.0", "requests>=2.13.0", "azure-identity>=1.5.0,<=1.6.0", "msal>=1.9.0,<2"],
```
yogilad, line 46: @AsafMah , We should probably allow all but major version changes

AsafMah: That's what it does for msal,
But for azure identity I don't understand the rationale
We either want:
1. All 1.5.*, and in that case we can leave it at ~=1.5.0
2. All 1.5.* and all of 1.6.*, and in this case we want >=1.5.0,<1.7.0
3. All of 1.*, and in this case >=1.5.0,<2

Allowing every version of 1.5.* but only 1.6.0 seems like a typo.

Ideally we would want option 3, but since they seem to do breaking changes in minors (https://pypi.org/project/azure-identity/), we probably should do 2

AsafMah:
Closing this - fixed in 2.3.1

Issue #332, opened 2021-07-21
RunTimeError: There is no current event loop


xneg:
Hello! We are creating client with this code:

```python
@staticmethod
def get_connection_string_builder(
        azure_ad_client_id, azure_ad_client_secret, azure_ad_tenant_id, cluster, msi, user_msi):
      # Managed Service Identity(MSI)
      if msi:
          return KustoConnectionStringBuilder.with_aad_managed_service_identity_authentication(
              cluster, client_id=user_msi
          )
      # Service Principal auth
      return KustoConnectionStringBuilder.with_aad_application_key_authentication(
          connection_string=cluster,
          aad_app_id=azure_ad_client_id,
          app_key=azure_ad_client_secret,
          authority_id=azure_ad_tenant_id,
      )

kcsb = self.get_connection_string_builder(
            azure_ad_client_id, azure_ad_client_secret, azure_ad_tenant_id, cluster, msi, user_msi
        )
self.kusto_client = KustoClient(kcsb)
```

#### Problem description
And after this changes https://github.com/Azure/azure-kusto-python/blob/77efe0608b66b50e947893fd0e53a36a22b6eab9/azure-kusto-data/azure/kusto/data/_token_providers.py#L69

We got an error:

```python
self.kusto_client = KustoClient(kcsb)
  File "/usr/local/lib/python3.7/site-packages/azure/kusto/data/client.py", line 763, in __init__
    self._auth_provider = _AadHelper(self._kcsb) if self._kcsb.aad_federated_security else None
  File "/usr/local/lib/python3.7/site-packages/azure/kusto/data/security.py", line 23, in __init__
    self.token_provider = ApplicationKeyTokenProvider(self.kusto_uri, kcsb.authority_id, kcsb.application_client_id, kcsb.application_key)
  File "/usr/local/lib/python3.7/site-packages/azure/kusto/data/_token_providers.py", line 485, in __init__
    super().__init__(kusto_uri)
  File "/usr/local/lib/python3.7/site-packages/azure/kusto/data/_token_providers.py", line 69, in __init__
    self._async_lock = asyncio.Lock()
  File "/usr/local/lib/python3.7/asyncio/locks.py", line 161, in __init__
    self._loop = events.get_event_loop()
  File "/usr/local/lib/python3.7/asyncio/events.py", line 644, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'Dummy-155'.
```

[...]

PR #335, opened 2021-07-22
Seperate Async and sync inside TokenProivders

AsafMah:
This PR adds a "is_async" parameter to separate the initialization and usage of TokenProvider.

This also includes an error when using an async operation on a sync provider and vice versa.

This solves #332, and it's now tested in test_basic_provider_in_thread

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -117,7 +117,7 @@ class TokenProviderBase(abc.ABC):
     def get_token(self):
         """Get a token silently from cache or authenticate if cached token is not found"""
         if self.is_async:
-            raise KustoAsyncUsageError("Method get_token can't be called on an asynchronous client")
+            raise KustoAsyncUsageError("get_token", self.is_async)
         self._init_once()
 
         token = self._get_token_from_cache_impl()
@@ -129,13 +129,13 @@ class TokenProviderBase(abc.ABC):
 
     def context(self) -> dict:
         if self.is_async:
-            raise KustoAsyncUsageError("Method context can't be called on an asynchronous client")
+            raise KustoAsyncUsageError("context", self.is_async)
         self._init_once(init_only_cloud=True)
         return self._context_impl()
 
     async def context_async(self) -> dict:
         if not self.is_async:
-            raise KustoAsyncUsageError("Method context_async can't be called on a synchronous client")
+            raise KustoAsyncUsageError("context_async", self.is_async)
 
         await self._init_once_async(init_only_cloud=True)
         return self._context_impl()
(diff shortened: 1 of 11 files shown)


azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -116,6 +116,8 @@ def _init_cloud(self):
 
     def get_token(self):
         """Get a token silently from cache or authenticate if cached token is not found"""
+        if self.is_async:
+            raise KustoAsyncUsageError("Method get_token can't be called on an asynchronous client")
```
ArielYehezkely, line 120: Use format and pass the method name as an argument

ArielYehezkely: maybe: ..called **from** an.. ?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -60,14 +60,14 @@ class TokenProviderBase(abc.ABC):
     _cloud_info = None
     _scopes = None
 
-    def __init__(self, kusto_uri: str):
+    def __init__(self, is_async: bool, kusto_uri: str):
```
ArielYehezkely, line 63: Consider add a default value so you won't need to pass 'False' it when using sync

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: The reasons I made it this way:
1. I prefer to keep it explicit, to minimize mistakes, especially since it's internals to it won't make customer's code more verbose
2. If it has a default value then it has to be at the end, right now it's the first argument in all of the token providers, so it's more consistent, and we won't have to change the code when adding parameters to child providers

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: WDYT of this way?

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: nit:
For this and all derived classes , not a fan on the position of this argument as the first.
It also conflicts with the order of this argument in AAD Helper

yogilad (dismissed):
Approved with a suggestion

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Where would you put in `MsiTokenProvider`
`def __init__(self, is_async: bool, kusto_uri: str, msi_args: dict = None)`?

yogilad: interesting.
I would either make it optional or use typing to state msi_args can be dict or none.
This is an internal class, so not critical

AsafMah:
I switched the order and made it optional, please take another look

yogilad (dismissed):

yogilad (approved):
please separate the version bump to a new PR

ArielYehezkely (approved):

ArielYehezkely (approved):

PR #339, opened 2021-08-03
Quick Start Sample App

yogilad:
#### Pull Request Description

Add a sample app which allows:
1. Quick Start for working with Python 
2. Template app for One Click


quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+    BlobDescriptor,
+    StreamDescriptor,
+    DataFormat,
+    ReportLevel,
+    IngestionMappingType,
+    KustoStreamingIngestClient,
+)
+
+# Todo - Config (Auto-Filled properties):
+kustoUri = "https://yogiladadx.westeurope.dev.kusto.windows.net"
```
danield137, line 27: if you want, we can read user input here

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
+# Todo - Start Here:
+#  1) Run setup.bat to install necessary dependencies.
```
danield137, line 2: I say just `pip install` to avoid OS specific files

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+tableName = "SampleTable"
+tableSchema = "SampleTable_schema"
+# Todo - Learn More: For additional information about supported data formats, see
+#  https://docs.microsoft.com/en-us/azure/data-explorer/ingestion-supported-formats
+fileFormat = DataFormat.CSV
+
+# Todo - Config (Optional): Change the authentication method from Device Code (User Prompt) to one of the other options
+#  Some of the auth modes require additional environment variables to be set in order to work (check the use below)
+#  Managed Identity Authentication only works when running as an Azure service (webapp, function, etc.)
+authenticationMode = "deviceCode"  # choose between: (deviceCode|managedIdentity|AppKey|AppCertificate)
```
danield137, line 42: there is also interactive auth which works pretty well

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+
+def die(error: str, ex: Exception = None):
+    print(error)
+    if ex is not None:
+        print(ex)
+
+    exit(-1)
+
+
+main()
```
danield137, line 260: usually, this is wrapped with `if __name__ == '__main__'` to ensure it's run as a script and not part of a module

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+    #  Create the clients once and use them as long as possible.
+    kusto_connection_string = create_connection_string(kustoUri, authenticationMode)
+    ingest_connection_string = create_connection_string(ingestUri, authenticationMode)
+    kusto_client = KustoClient(kusto_connection_string)
+    ingest_client = QueuedIngestClient(ingest_connection_string)
+
+    # Todo - Learn More: For additional information on how to create tables see: https://docs.microsoft.com/en-us/azure/data-explorer/one-click-table
+    print("")
+    print(f"Creating table '{databaseName}.{tableName}' if needed:")
+    # Todo (Yochai) what's the table creation command
```
danield137, line 65: `.create table Foo (col1: int, col2: string) ifnotexists`

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+    # Todo - Tip: Avoid creating a new Kusto Client for each use.
+    #  Create the clients once and use them as long as possible.
+    kusto_connection_string = create_connection_string(kustoUri, authenticationMode)
+    ingest_connection_string = create_connection_string(ingestUri, authenticationMode)
+    kusto_client = KustoClient(kusto_connection_string)
+    ingest_client = QueuedIngestClient(ingest_connection_string)
+
+    # Todo - Learn More: For additional information on how to create tables see: https://docs.microsoft.com/en-us/azure/data-explorer/one-click-table
+    print("")
+    print(f"Creating table '{databaseName}.{tableName}' if needed:")
```
danield137, line 64: if not exists is more correct I guess

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+    # Todo (Yochai) what's the table creation command
+    command = ".show version"
+    if not run_control_command(kusto_client, databaseName, command):
+        print("Failed to create or validate table exists.")
+        exit(-1)
+
+    # Todo - Learn More: For additional information on Kusto Query Language see: https://docs.microsoft.com/en-us/azure/data-explorer/write-queries
+    print("")
+    print(f"Initial row count for '{databaseName}.{tableName}' is:")
```
danield137, line 73: why is this important?

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+    #   1) more then a 1000 files were queued for ingestion for the same table by the same user
+    #   2) more then 1GB of data was queued for ingestion for the same table by the same user
+    #   3) More then 5 minutes have passed since the first file was queued for ingestion for the same table by the same user
+    #  In order to speed up this sample app, we attempt to modify the default ingestion policy to ingest data after 10 seconds have passed
+    #  For additional information on customizing the ingestion batching policy see:
+    #   https://docs.microsoft.com/en-us/azure/data-explorer/kusto/management/batchingpolicy
+    #  You may also skip the batching for some files using the FlushImmediatly property, though this option should be used with care as it is inefficient
+    print("")
+    print(f"Altering the batching policy for '{tableName}'")
+    batching_policy = '{ "MaximumBatchingTimeSpan": "00:00:10", "MaximumNumberOfItems": 500, "MaximumRawDataSizeMB": 1024 }'
```
danield137, line 87: might as well have maximum items = 1

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+        print("Failed to alter the ingestion policy!")
+        print("This could be the result of insufficient permissions.")
+        print("The sample will still run, though ingestion will be delayed for 5 minutes")
+
+    # Todo - Learn More: For additional information on how to ingest data to Kusto in Python see:
+    #  https://docs.microsoft.com/en-us/azure/data-explorer/python-ingest-data
+    print("")
+    ingest_file = input("Please enter a directory of files to ingest from:")
+    print(f"Attempting to ingest '{ingest_file}'")
+    ingest_data_from_file(ingest_client, databaseName, tableName, ingest_file)
```
danield137, line 99: this requires user intervention... should we just ingest a sample? (one that works against the table created about)

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+
+    print("")
+    print("Sleeping for a few seconds to make sure queued ingestion has completed")
+    print("Mind, this may take longer dependeing on the file size and ingestion policy")
+    time.sleep(20)
```
danield137, line 104: probably better to sleep for less and out put something to the screen.. 20 seconds can feel like a long time :)

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+
+    print("")
+    print(f"Post ingestion row count for '{databaseName}.{tableName}' is:")
+    run_query(kusto_client, databaseName, f"{tableName} | summarize count()")
```
danield137, line 108: maybe also `limit 1`

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,261 @@
...
+    # Todo - Tip: for optimal ingestion batching it's best to specify the uncompressed data size in the file descriptor
+    file_descriptor = FileDescriptor(f"{file_path}")
+    client.ingest_from_file(file_descriptor, ingestion_properties=ingestion_props)
+
+    # Todo - Tip: Kusto can also ingest data from blobs, open streams and pandas dataframes.
+    #  See the python SDK azure.kusto.ingest samples for additional references.
+
+
+def create_connection_string(cluster: str, auth_mode: str) -> KustoConnectionStringBuilder:
+    if auth_mode == "deviceCode":
```
danield137, line 177: this kinda works against our static functions approach. if a user were to copy-paste this...

quick_start/kusto_sample_app.py
yogilad: The idea is for this to include place holders which will later be filled automatically by one click.
I would rather limit the amount of user duct-tape code, so that this sample would include mostly code which can be quickly adapted for the dev needs.

quick_start/kusto_sample_app.py
yogilad: Awesome. Forgot about this

quick_start/kusto_sample_app.py
yogilad: It's not. Just an excuse to demo a simple query...

quick_start/kusto_sample_app.py
yogilad: I would rather not steer them in a direction of a policy which I would not recommend.
There's also some info on ingestion batching to help them understand which direction they should actually take.

quick_start/kusto_sample_app.py
yogilad: This is supposed to integrate with one-click so the table schema will not be fixed.
In theory we can ingest the sample provided for oneClickm but that would mean the same data is ingested in each run, so not sure if this is a good idea.

quick_start/kusto_sample_app.py
yogilad: Less time usually results in the next query not showing the update.

quick_start/kusto_sample_app.py
yogilad: Not sure what you mean.
Summarize count() produces one line.

quick_start/kusto_sample_app.py
yogilad: Explain?

quick_start/kusto_sample_app.py
danield137: oh no, I meant to have a loop like:
```
for _ in range(20):
   time.sleep(1)
   print('.')
```

anyway, just a suggestion

quick_start/kusto_sample_app.py
danield137: yes, but it doesn't show an example output. it is quite common to do a `head` operation to guarantee data validity.

quick_start/kusto_sample_app.py
danield137: `KustoConnectionStringBuilder.with_method` should be clear and explicit enough. have a wrapper function might steer users to doing the same, which would make our static function obsolete. It is not a bad thing per se, just a comment.

danield137 (approved):
Looks good! Clear and verbose.

quick_start/kusto_sample_app.py
yogilad: Oh, that's just cause the script demonstrates multiple auth modes (certificates, keys, etc.).
You need some way of selecting between them and I did not want to push this code into the main function.

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,334 @@
...
+    command = f".create table {tableName} {tableSchema}"
+    if not run_control_command(kusto_client, databaseName, command):
+        print("Failed to create or validate table exists.")
+        exit(-1)
+
+    wait_for_user()
+
+    print("")
+    print(f"Altering the batching policy for '{tableName}'")
+    # Todo - Tip: this is commonly a one-time command
```
ohadbitt: Specify this is for testing only

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,334 @@
...
+    wait_for_user()
+
+    # Todo - Learn More: For additional information on Kusto Query Language see: https://docs.microsoft.com/en-us/azure/data-explorer/write-queries
+    print("")
+    print(f"Initial row count for '{databaseName}.{tableName}' is:")
+    run_query(kusto_client, databaseName, f"{tableName} | summarize count()")
+    wait_for_user()
+
+    # Todo - Learn More: For additional information on how to ingest data to Kusto in Python see:
+    #  https://docs.microsoft.com/en-us/azure/data-explorer/python-ingest-data
```
ohadbitt, line 106: drop the "en-us" from all links

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,334 @@
...
+        #  https://docs.microsoft.com/en-us/azure/data-explorer/ingestion-supported-formats
+        data_format=file_format,
+        # Todo - Config: Setting the ingestion batching policy takes up to 5 minutes to have an effect.
+        #  For the sake of the sample we set Flush-Immediately, but in practice it should not be commonly used.
+        #  Comment the below line after running the sample for the first few times!
+        flush_immediately=True,
+        ingestion_mapping_reference=mapping_ref,
+    )
+
+    # Todo - Tip: for optimal ingestion batching it's best to specify the uncompressed data size in the file descriptor
```
ohadbitt: then lets show how to specify it

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,334 @@
+# Todo - Start Here:
```
ohadbitt: why is it called Todo?

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,334 @@
...
+
+        # Todo - learn more:  for more information about providing inline mappings or mapping references see:
+        #  https://docs.microsoft.com/en-us/azure/data-explorer/kusto/management/mappings
+
+        wait_for_user()
+
+        if mapping_exists:
+            print("")
+            print(f"Attempting to ingest '{jsonSample}'")
+            ingest_data_from_file(ingest_client, databaseName, tableName, jsonSample, DataFormat.MULTIJSON, jsonMappingRef)
```
ohadbitt: add some comment to suggest to always use MULTIJSON when ingesting Json

quick_start/kusto_sample_app.py
```diff
@@ -0,0 +1,334 @@
...
+            wait_for_user()
+
+    print("")
+    print("Sleeping for a few seconds to make sure queued ingestion has completed")
+    print("Mind, this may take longer depending on the file size and ingestion policy")
+    for x in range(20, 0, -1):
+        print(f"{x} ", end="\r")
```
ohadbitt, line 138: did you mean 

`end =""`

 ?

yogilad: No, I specifically want it to be carriage return, so that the next line will print over the current one.

quick_start/kusto_sample_app.py
yogilad: There's a whole section explaining it in line 88 on the latest version.
LMK if you think it needs more info

quick_start/kusto_sample_app.py
yogilad: There are specific comments where I want the user to intervene in the code by doing something.
Plus some IDEs highlight those differently.
Note the latest version has far few of those

quick_start/kusto_sample_app.py
yogilad: Good?
```
# Tip: When ingesting json files, if a each row is represented by a single line json, use MULTIJSON format even if the file only includes one line.
# When the json contains whitespace formatting, use SINGLEJSON. IN this case only one row/json per file is allowed.
```

PR #343, opened 2021-10-28
Support not passing blobs size

AsafMah:
Added support for not passing blobs size


azure-kusto-ingest/azure/kusto/ingest/descriptors.py
```diff
@@ -31,18 +31,18 @@ class FileDescriptor:
...
         :type size: int.
-        :param source_id uuid: a v4 uuid to serve as the sources id.
-        :type source_id: str (of a uuid4) or uuid4.
+        :param source_id: a v4 uuid to serve as the sources id.
+        :type source_id: OptionalUUID
         """
         self.path = path
         self._size = size
-        self._detect_size_once = size < 1
+        self._detect_size_once = not size
```
yogilad, line 45: That's a breaking change as 0 size (the previous default) would be treated as no detection.
Why not make the blob default 0 instead of None?

AsafMah: not 0 == True

The behavior doesn't change

yogilad (approved):

PR #349, opened 2021-12-01
Fix pandas types

ArielYehezkely:
#### Pull Request Description

Fix conversion of numeric types when using pandas

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Fix numeric type conversion when using pandas data frames

### DIFF
--- a/azure-kusto-data/azure/kusto/data/helpers.py
+++ b/azure-kusto-data/azure/kusto/data/helpers.py
@@ -55,11 +55,11 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
         elif col.column_type == "int" or col.column_type == "long":
             frame[col.column_name] = frame[col.column_name].astype("Int64")
         elif col.column_type == "real" or col.column_type == "decimal":
-            frame[col.column_name] = frame[col.column_name].replace('NaN', np.NaN)
-            frame[col.column_name] = frame[col.column_name].replace('Infinity', np.PINF)
-            frame[col.column_name] = frame[col.column_name].replace('-Infinity', np.NINF)
-            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors='coerce')
-            frame[col.column_name] = frame[col.column_name].astype('Float64')
+            frame[col.column_name] = frame[col.column_name].replace("NaN", np.NaN)
+            frame[col.column_name] = frame[col.column_name].replace("Infinity", np.PINF)
+            frame[col.column_name] = frame[col.column_name].replace("-Infinity", np.NINF)
+            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors="coerce")
+            frame[col.column_name] = frame[col.column_name].astype("Float64")
         elif col.column_type == "datetime":
             frame[col.column_name] = pd.to_datetime(frame[col.column_name])
         elif col.column_type == "timespan":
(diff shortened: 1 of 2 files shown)


azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -50,9 +52,17 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
         if col.column_type == "bool":
             frame[col.column_name] = frame[col.column_name].astype(bool)
-        if col.column_type == "datetime":
+        elif col.column_type == "int" or col.column_type == "long":
+            frame[col.column_name] = frame[col.column_name].astype("Int64")
+        elif col.column_type == "real" or col.column_type == "decimal":
+            frame[col.column_name] = frame[col.column_name].replace('NaN', np.NaN)
+            frame[col.column_name] = frame[col.column_name].replace('Infinity', np.PINF)
+            frame[col.column_name] = frame[col.column_name].replace('-Infinity', np.NINF)
+            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors= 'coerce')
```
janlugt, line 61: I think the tests are failing because of a formatting issue here. The space after errors= should be removed.
```suggestion
            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors='coerce')
```

janlugt (dismissed):
LGTM

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -50,9 +52,17 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
             frame[col.column_name] = frame[col.column_name].astype(bool)
-        if col.column_type == "datetime":
+        elif col.column_type == "int" or col.column_type == "long":
+            frame[col.column_name] = frame[col.column_name].astype("Int64")
+        elif col.column_type == "real" or col.column_type == "decimal":
+            frame[col.column_name] = frame[col.column_name].replace("NaN", np.NaN)
+            frame[col.column_name] = frame[col.column_name].replace("Infinity", np.PINF)
+            frame[col.column_name] = frame[col.column_name].replace("-Infinity", np.NINF)
+            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors="coerce")
+            frame[col.column_name] = frame[col.column_name].astype("Float64")
```
AsafMah, line 62:
```suggestion
            frame[col.column_name] = frame[col.column_name].replace("NaN", np.NaN).replace("Infinity", np.PINF).replace("-Infinity", np.NINF)
            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors="coerce").astype("Float64")
```

You can reduce the amount of assignments, it's clearer IMO

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -50,9 +52,17 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
     for col in table.columns:
         if col.column_type == "bool":
             frame[col.column_name] = frame[col.column_name].astype(bool)
-        if col.column_type == "datetime":
+        elif col.column_type == "int" or col.column_type == "long":
+            frame[col.column_name] = frame[col.column_name].astype("Int64")
+        elif col.column_type == "real" or col.column_type == "decimal":
```
AsafMah, line 57: Should we also check for "double" since it's an alias (https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/scalar-data-types/)

(I guess we don't check for "date" and "time" either)

AsafMah (dismissed):
Looks good overall

azure-kusto-data/azure/kusto/data/helpers.py
ArielYehezkely: The alias is for user comfort, the type underneath will be "real" and the column type in the response will be "real" even if the user defined it as "double".

PR #354, opened 2021-12-21
Enhance DataFormat

yihezkel:
#### Pull Request Description

Note: I didn't realize (until the merge conflicts now) that Asaf recently made some changes to is_mapping_required and is_binary/is_compressible. I don't think his PR affects any of my changes here, except for a decision regarding whether we want is_binary or is_compressible. I'm open to either, but prefer the former if we have any reason to care whether a format is binary other than to determine if it's compressible, and otherwise prefer the latter.

I will discuss this with relevant people.

---

#### Future Release Comment
**Breaking Changes:**
- Enhance DataFormat to include relevant properties, instead of maintaining lists of which values satisfy those properties externally.


azure-kusto-data/azure/kusto/data/data_format.py
```diff
@@ -2,22 +2,37 @@
...
+    PARQUET = ("parquet", IngestionMappingType.PARQUET, False, False)
+    ORC = ("orc", IngestionMappingType.ORC, False, False)
+    RAW = ("raw", IngestionMappingType.UNKNOWN, False, True)
+    W3CLOGFILE = ("w3clogfile", IngestionMappingType.W3CLOGFILE, False, True)
+
+    def __init__(self, value_: str, ingestion_mapping_type: IngestionMappingType, mapping_required: bool, compressible: bool):
+        self.value_ = value_ # Formatted how Kusto Service expects it
+        self.ingestion_mapping_type = ingestion_mapping_type
+        self.mapping_required = mapping_required
+        self.compressible = compressible # Binary formats should not be compressed
```
yihezkel, line 38: I didn't realize (until the merge conflicts now) that Asaf recently made some changes to is_mapping_required and is_binary/is_compressible. I don't think his PR affects any of my changes here, except for a decision regarding whether we want is_binary or is_compressible. I'm open to either, but prefer the former if we have any reason to care whether a format is binary other than to determine if it's compressible, and otherwise prefer the latter.

I will discuss this with relevant people.

azure-kusto-data/azure/kusto/data/data_format.py
```diff
@@ -2,22 +2,37 @@
...
+    SINGLEJSON = ("singlejson", IngestionMappingType.JSON, True, True)
+    MULTIJSON = ("multijson", IngestionMappingType.JSON, True, True)
+    AVRO = ("avro", IngestionMappingType.AVRO, True, True)
+    PARQUET = ("parquet", IngestionMappingType.PARQUET, False, False)
+    ORC = ("orc", IngestionMappingType.ORC, False, False)
+    RAW = ("raw", IngestionMappingType.UNKNOWN, False, True)
+    W3CLOGFILE = ("w3clogfile", IngestionMappingType.W3CLOGFILE, False, True)
+
+    def __init__(self, value_: str, ingestion_mapping_type: IngestionMappingType, mapping_required: bool, compressible: bool):
+        self.value_ = value_ # Formatted how Kusto Service expects it
```
yihezkel, line 35: I also considered calling this `kusto_value` or `service_value` or `service_formatted_value` (in order of my preference).

Thoughts?

azure-kusto-ingest/tests/sample.py
```diff
@@ -10,12 +10,13 @@
     FileDescriptor,
     BlobDescriptor,
     StreamDescriptor,
-    DataFormat,
     KustoStreamingIngestClient,
     ManagedStreamingIngestClient,
     IngestionStatus,
 )
 
+from azure.kusto.data.data_format import DataFormat
```
AsafMah, line 18: Put it next to the other azure.kusto.data imports

In fact, shouldn't we just export it from azure.kusto.data directly?

azure-kusto-ingest/tests/test_ingestion_blob_info.py
```diff
@@ -19,6 +18,8 @@
     ValidationImplications,
 )
 
+from azure.kusto.data.data_format import DataFormat
```
AsafMah, line 21: External imports should be put before internal imports (https://www.python.org/dev/peps/pep-0008/#imports)

azure-kusto-data/azure/kusto/data/data_format.py
AsafMah: I don't like value_, I like your suggestions (in the same order :) )

azure-kusto-data/azure/kusto/data/data_format.py
AsafMah: is_binary was only used for checking if to compress.  
I can't really think of another reason to want this information, so I'm fine with just is_compressible

AsafMah (changes requested):
small comments

AsafMah:
One thing that still bothers me, is that DataFormat is exported from azure.kusto.ingest directly, but in data it's from azure.kusto.data.data_format.

This before was because we had two of them.
I think the correct thing is to not export it from ingest, and users will import it from data like they have to do with the connection string for example.

On the other hand it's a breaking change so I'm not sure.

yihezkel:
@AsafMah (can't reply to your comment, presumably since it's not on code)

I suspect the 2 imports of data_format from the ingest package (base_ingest_client and test_managed_streaming_ingest) were because of my recent merge with master (managed streaming), where you had included in the ingest package imports of data_format from ingest (because that's where they were when you did your PR). This PR moves data_format to data, so as part of that merge I should've changed those imports. I just did that now.

As for the export of data_format in ingest's __init__, I kept it there to not break anything. OTOH, we're intending to do a major version bump soon anyway (see the recently merged managed streaming PR, referenced above). So I removed it from there as well and will have to document that in the release.

AsafMah (approved):

Issue #355, opened 2021-12-21
"dataframe_from_result_table" fails to handle out-of-range datetimes.


liamkirton:
Kusto datetimes start from "0001-01-01 00:00:00.0", whereas Pandas start from "1677-09-21 00:12:43.145224193". As a consequence, dataframe_from_result_table raises an Exception when Kusto returns out-of-range datetimes.

The following KQL will reproduce:

```
let T = datatable(a:datetime, b:string) [
    datetime(0001-01-01), "foo"
];
T | take 1;
```

The fix should be simply adding `errors="coerce"` to the datetime conversion in dataframe_from_result_table.

yogilad:
@ArielYehezkely , @AsafMah , do you see any objection to the change?

If not, @liamkirton could you please add a test for the change and we'll accept the contribution.
Thanks!

liamkirton:
Thanks @yogilad - I've updated with PR and added `tests/test_helpers.py `, to exercise `dataframe_from_result_table`. Really happy to address any additional comments!

yogilad:
Added @AsafMah & @ArielYehezkely  to handle review

AsafMah:
Merged

PR #356, opened 2021-12-21
Handle invalid datetimes in dataframe_from_result_table

liamkirton:
This is a fix for #355, where dataframe_from_result_table fails to handle out-of-range datetime conversion.


AsafMah:
Looks good, can you add a test?

azure-kusto-data/tests/test_helpers.py
```diff
@@ -0,0 +1,265 @@
...
+    import pandas
+    import numpy
+
+    PANDAS = True
+except:
+    pass
+
+
+# Sample response containing conversion edge cases.
+RESPONSE_TEXT = """
```
AsafMah, line 22: Move this to a json file like we do in other tests

azure-kusto-data/tests/test_helpers.py
```diff
@@ -0,0 +1,265 @@
...
+},
+{
+    "FrameType": "DataSetCompletion",
+    "HasErrors": false,
+    "Cancelled": false
+}]
+"""
+
+
+class TestDataFrameFromResultsTable(unittest.TestCase):
```
AsafMah, line 213: We use pytest, so please remove the inheritance from unittest.TestCase, as it may interfere

AsafMah (commented):
Looks good, just some small changes

azure-kusto-data/tests/test_helpers.py
liamkirton: Done!

azure-kusto-data/tests/test_helpers.py
liamkirton: Done!

AsafMah:
Can you see why the test fails?

liamkirton:
The test failure on 3.6 was due to differences in Pandas behaviour: between the version on 3.6 and 3.7, the minimum representable datetime changes slightly...

pd._libs.tslibs.timestamps.Timestamp.min == Timestamp('1677-09-21 00:12:43.145225')
vs.
pd._libs.tslibs.timestamps.Timestamp.min == Timestamp('1677-09-21 00:12:43.145224193')

I've relaxed the test criteria to just check that parsing a close-to-min/max timestamp succeeds.

AsafMah (approved):

PR #357, opened 2021-12-22
Cloud Info improvements

AsafMah:
**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
-Don't fetch cloudinfo if not needed
-Pass correct scopes to az cli and msi

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -67,7 +67,7 @@ class TokenProviderBase(abc.ABC):
         else:
             self._lock = Lock()
 
-    def _init_once(self, init_only_cloud=False):
+    def _init_once(self, init_only_resources=False):
         if self._initialized:
             return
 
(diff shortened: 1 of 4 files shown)


yogilad (dismissed):

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -77,9 +75,9 @@ def _init_once(self, init_only_cloud=False):
             if self._initialized:
                 return
 
-            if not self._cloud_initialized:
-                self._init_cloud()
-                self._cloud_initialized = True
+            if not self._resources_initialized:
+                self._init_resources()
```
yihezkel, line 79: What's the benefit of `_resources_initialized` and `_init_resources` here in `TokenProviderBase`?

They seem to only do something in `CloudInfoTokenProvider`.

AsafMah: The problem is that it's called in the middle of the method, and since I don't want to re-write the entire __init__ in the subclass, the virtual method works ok.

I'd happy to hear if you have a cleaner solution

yihezkel (dismissed):

yihezkel (approved):

PR #358, opened 2021-12-23
Make sure uuids are always generated, and are always the correct type

AsafMah:
Also more type annotations


yihezkel (approved):
nit: I would just add a test without providing a UUID, and verify that one is created.

AsafMah:
@yihezkel I added some unit tests LMKWYT

Also - do you think that for BlobDescriptor we should add a default value None to size, to keep parity with the ctors of StreamDescriptor and FileDescriptor ?

yihezkel:
@AsafMah Good call, yeah I think so.

I'm not entirely clear why we don't also attempt to detect its size if BlobDescriptor's size is None (Java too). I get that a remote op is more expensive than on the local filesystem, but if we don't provide a size, the service will have to download it anyway in `Kusto.DataManagement.Common.BlobUncompressedSizeCalculator.Calculate`.

AsafMah:
@yihezkel I'll add it to the doc, but I think we generally prefer the sdk to be lightweight and give the job to the server.

For now will you approve the PR?

yihezkel (approved):

PR #359, opened 2021-12-23
Streaming ingest mapping error

AsafMah:
Raise error when using streaming ingest with a type that needs mapping reference without a mapping reference


azure-kusto-ingest/azure/kusto/ingest/streaming_ingest_client.py
```diff
@@ -45,6 +46,11 @@ def ingest_from_stream(self, stream_descriptor: Union[StreamDescriptor, IO[AnySt
     def _ingest_from_stream_with_client_request_id(
         self, stream_descriptor: Union[StreamDescriptor, IO[AnyStr]], ingestion_properties: IngestionProperties, client_request_id: Optional[str]
     ) -> IngestionResult:
+
+        # We need this check in addition to the one in the IngestionProperties constructor since streaming ingestion only supports mapping reference
```
yihezkel, line 50: Do we still want to add this at this point? I suspect we don't want to add any new restrictions on mappings which you determined aren't required.

PR #360, opened 2021-12-28
Proxy support

AsafMah:
Adds proxy support to query clients 

Testing - tested manually and it works.

Couldn't manage to set up an e2e, contributions are welcome


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -765,24 +766,26 @@ class KustoClient(_KustoClientBase):
     # The maximum amount of connections to be able to operate in parallel
     _max_pool_size = 100
 
-    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
+    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str], proxies: Optional[Dict[str, str]] = None):
```
ArielYehezkely, line 769: Do we have to pass it in the C'tor? I know it's optional but also an uncommon use case..

AsafMah: It's the easiest way to ensure it's in propagated all the way down

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/client.py
ArielYehezkely: You can add it as a property and override the necessary members when the user sets it - but your call

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -232,7 +233,7 @@ def _context_impl(self) -> dict:
     def _init_impl(self):
         pass
 
-    def _get_token_impl(self) -> dict:
+    def _get_token_impl(self) -> None:
```
yogilad, line 236: Instead if this make the class abstract and don't implement the abstract API?

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -307,7 +308,7 @@ def _init_impl(self):
     def _get_token_impl(self) -> dict:
         try:
             if self._msi_auth_context is None:
-                self._msi_auth_context = ManagedIdentityCredential(**self._msi_args)
+                self._msi_auth_context = ManagedIdentityCredential(**self._msi_args, proxies=self._proxies)
```
yogilad, line 311: That does not make sense.
MSI is A) Always in Azure, B) Always on a link local IP for your current node.
Can't say if this is redundant or a security issue, but let's not allow this

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -319,7 +320,7 @@ def _get_token_impl(self) -> dict:
     async def _get_token_impl_async(self) -> Optional[dict]:
         try:
             if self._msi_auth_context_async is None:
-                self._msi_auth_context_async = AsyncManagedIdentityCredential(**self._msi_args)
+                self._msi_auth_context_async = AsyncManagedIdentityCredential(**self._msi_args, proxies=self._proxies)
```
yogilad, line 323: Same here

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -20,10 +20,18 @@
...
 class KustoClient(_KustoClientBase):
     @documented_by(KustoClientSync.__init__)
-    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
-        super().__init__(kcsb)
+    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str], proxies: Optional[Dict[str, str]] = None):
+        super().__init__(kcsb, proxies)
+
+        if proxies is not None:
+            if len(proxies) > 1:
+                raise KustoClientError("Only one proxy is supported in async client")
```
yogilad, line 28: Why are you enforcing this?
Either let the underlying clients decide what's right and wrong or ask for a single proxy an build the dictionary yourself.
I vote on the former

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -20,10 +20,18 @@
...
+            self._proxy = list(proxies.values())[0]
```
yogilad, line 29: Ha?

azure-kusto-data/azure/kusto/data/security.py
```diff
@@ -23,14 +24,16 @@ class _AadHelper:
...
         if kcsb.interactive_login:
-            self.token_provider = InteractiveLoginTokenProvider(self.kusto_uri, kcsb.authority_id, kcsb.login_hint, kcsb.domain_hint, is_async=is_async)
+            self.token_provider = InteractiveLoginTokenProvider(
+                self.kusto_uri, kcsb.authority_id, kcsb.login_hint, kcsb.domain_hint, is_async=is_async, proxies=proxies
+            )
         elif all([kcsb.aad_user_id, kcsb.password]):
-            self.token_provider = UserPassTokenProvider(self.kusto_uri, kcsb.authority_id, kcsb.aad_user_id, kcsb.password, is_async=is_async)
+            self.token_provider = UserPassTokenProvider(self.kusto_uri, kcsb.authority_id, kcsb.aad_user_id, kcsb.password, is_async=is_async, proxies=proxies)
         elif all([kcsb.application_client_id, kcsb.application_key]):
             self.token_provider = ApplicationKeyTokenProvider(
```
yogilad, line 38: what about this guy?

yogilad: Also ApplicationCertificateTokenProvider

yogilad (changes requested):

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: The base class is abstract.
I think not implementing an abstract method is an anti-pattern.
Explicit is better than implicit

azure-kusto-data/azure/kusto/data/aio/client.py
AsafMah: The thing is `requests` and almost everything else accepts a dictionary, but `aiohttp` accepts a single string.
The two options are either accept a dictionary for the async client, and pass it on to everything else, and for aiohttp directly convert it to string (that's what I did here, and that's why there is this check and the next comment).

The alternative is to accept a string for the async client, convert it to a dictionary and pass it forward.
I chose the first one because it keeps the interface the same between the sync/async clients.

So do you understand now? WDYT?

azure-kusto-data/azure/kusto/data/_token_providers.py
yogilad: In that case do not change the interface (or change everything to optional[dict]

azure-kusto-data/azure/kusto/data/aio/client.py
yogilad: I think the idea is that when your are using aiohttp you are already bound to a specific address which you need to select from the dictionary (or None if it is not there).
Does it make sense?

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Is it really changing the interface?
The base class has a weaker type, and this change makes it into a stronger type.
It still fits in a "Optional[dict]".
There are impls that only return dict, I don't think changing them to Optional[dict] will help

azure-kusto-data/azure/kusto/data/aio/client.py
AsafMah: Sorry I don't understand

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: Everything now returns Optional[Dict]

azure-kusto-data/azure/kusto/data/aio/client.py
AsafMah: Solved offline

azure-kusto-data/azure/kusto/data/client.py
AsafMah: Ok I thought about it and I think you're right.
I changed it, let me know what you think

ArielYehezkely (approved):

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -58,9 +62,9 @@ class TokenProviderBase(abc.ABC):
     _resources_initialized: bool = False
     _kusto_uri: str
```
yihezkel, line 63: Remove

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -470,7 +482,15 @@ def _get_token_from_cache_impl(self) -> dict:
...
 
-    def __init__(self, kusto_uri: str, authority_id: str, login_hint: Optional[str] = None, domain_hint: Optional[str] = None, is_async: bool = False):
+    def __init__(
+        self,
+        kusto_uri: str,
+        authority_id: str,
+        login_hint: Optional[str] = None,
+        domain_hint: Optional[str] = None,
+        is_async: bool = False,
+        proxy_dict: Optional[dict] = None,
```
yihezkel, line 492: Inconsistency- get the proxy in constructor, but don't use it.
Check elsewhere as well for same issue.

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -541,7 +563,17 @@ class ApplicationCertificateTokenProvider(CloudInfoTokenProvider):
...
+    def __init__(
+        self,
+        kusto_uri: str,
+        client_id: str,
+        authority_id: str,
+        private_cert: str,
+        thumbprint: str,
+        public_cert: str = None,
+        is_async: bool = False,
+        proxy_dict: Optional[dict] = None,
```
yihezkel, line 575: Same issue

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -1,962 +1,975 @@
...
+
+
+class _KustoClientBase(abc.ABC):
+    API_VERSION = "2019-02-13"
+
+    _mgmt_default_timeout = timedelta(hours=1, seconds=30)
+    _query_default_timeout = timedelta(minutes=4, seconds=30)
+    _streaming_ingest_default_timeout = timedelta(minutes=10)
+
+    _auth_provider: _AadHelper
```
yihezkel, line 694: Naming + no need in both this and child class, have child pass to here

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -1,962 +1,975 @@
...
+        self._query_endpoint = "{0}/v2/rest/query".format(self._kusto_cluster)
+        self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(self._kusto_cluster)
+        self._request_headers = {
+            "Accept": "application/json",
+            "Accept-Encoding": "gzip,deflate",
+            "x-ms-client-version": "Kusto.Python.Client:" + VERSION,
+            "x-ms-version": self.API_VERSION,
+        }
+
+    def set_proxy(self, proxy: str):
```
yihezkel, line 715: proxy -> proxyUrl

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -1,962 +1,975 @@
...
+        self._proxy = proxy
+        self._proxy_dict = {"http": proxy, "https": proxy}
+        self._auth_provider.token_provider.set_proxy_dict(self._proxy_dict)
```
yihezkel, line 718: set_proxy_dict with no args defaults proxy_dict to "{"http": proxy, "https": proxy}"

yihezkel: And remove proxy_dict from this (and any similar) classes

yihezkel (approved):

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -30,6 +30,10 @@ class AsyncManagedIdentityCredential:
         def __init__(self):
             raise KustoAioSyntaxError()
 
+    class AsyncAzureCliCredential:
+        def __init__(self):
+            raise KustoAioSyntaxError()
```
yihezkel, line 35: nit: Add a comment since I needed you to explain this to me when w ereviewed it together

AsafMah: Added, WDYT?

yogilad (dismissed):

azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -197,17 +200,22 @@ def _valid_token_or_throw(self, token: dict, context: str = "") -> dict:
 
         return token
 
+    def set_proxy(self, proxy_url: str):
```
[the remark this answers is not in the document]
yogilad: ignore this - we don't use any other protocol so it does not matter

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -1,962 +1,972 @@
...
+    _aad_helper: _AadHelper
+
+    def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str], is_async):
+        self._kcsb = kcsb
+        self._proxy_url: Optional[str] = None
+        if not isinstance(kcsb, KustoConnectionStringBuilder):
+            self._kcsb = KustoConnectionStringBuilder(kcsb)
+        self._kusto_cluster = self._kcsb.data_source
+
+        # notice that in this context, federated actually just stands for add auth, not aad federated auth (legacy code)
```
yihezkel, line 702: nit: add -> aad

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -1,962 +1,972 @@
...
+        self._streaming_ingest_endpoint = "{0}/v1/rest/ingest/".format(self._kusto_cluster)
+        self._request_headers = {
+            "Accept": "application/json",
+            "Accept-Encoding": "gzip,deflate",
+            "x-ms-client-version": "Kusto.Python.Client:" + VERSION,
+            "x-ms-version": self.API_VERSION,
+        }
+
+    def set_proxy(self, proxy_url: str):
+        self._proxy_url = proxy_url
```
yihezkel, line 717: If we use the `aad_helper` to maintain the `proxy_dict`, why not use it for the `proxy_url`?

I think it would be cleaner to remove the `_proxy_url` from the client, and instead the `aad_helper` would maintain the `proxy_url`. I'd then remove the `proxy_dict` from being a property of `aad_helper`, and instead a method that dynamically returns the `proxy_dict` based on its `proxy_url` property.

yihezkel: Sorry, my mistake - it's not in aad_helper but in the token provider.

yihezkel (approved):

PR #361, opened 2021-12-28
Update minimum python to 3.7

AsafMah:
-Update minimum python to 3.7
-Test 3.10
-Pipeline improvements - add unit test summary
-Pipeline improvements - use official black reporting


azure-kusto-data/setup.py
```diff
@@ -34,9 +34,10 @@
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",
         "Topic :: Software Development",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
yihezkel, line 40: I'm not familiar with setup.py, but why are we explicitly listing versions beyond the minimum? Does this prevent using version 3.11 once it comes out until these 2 setup.pys are manually upgraded?

AsafMah: These are just tags that you can search with in pypi.
So if the user for example searches for packages with the "3.6" classifier they will not find us.

It does mean we need to keep it updated.

yihezkel: Is there no way to dynamically indicate we support v3.6+ without having to manually update?

AsafMah: Not that I found (and we support 3.7+)

yihezkel:
Note that the build isn't reporting success.

AsafMah:
@yihezkel 
Ok, i changed the gates to match this.
Now we need it merged because the others won't pass without it (testing on 3.10)

yihezkel (approved):

PR #363, opened 2022-01-10
IngestionProperties Improvements

AsafMah:
- Renamed IngestionMappingType to IngestionMappingKind to align with other SDKs

- Added advanced checks for column mappings, see https://github.com/Azure/azure-kusto-java/pull/205
- Added missing formats and mappings
- Renamed "ingestion_mapping" parameter to "column_mappings" for clarity and alignment

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
@@ -27,14 +27,6 @@ class KustoMissingMappingError(KustoClientError):
     or when provided a mapping kind without a mapping.
     """
 
-    @staticmethod
-    def required_data_format(data_format: DataFormat) -> "KustoMissingMappingError":
-        return KustoMissingMappingError(f"When stream format is '{data_format.kusto_value}', a mapping must be provided.")
-
-    @staticmethod
-    def required_with_kind(kind: IngestionMappingKind) -> "KustoMissingMappingError":
-        return KustoMissingMappingError(f"When ingestion mapping kind is set ('{kind.value}'), a mapping must be provided.")
-
 
 class KustoInvalidEndpointError(KustoClientError):
     """Raised when trying to ingest to invalid cluster type."""
--- a/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/exceptions.py
@@ -1,39 +1,39 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
-
+from azure.kusto.data.data_format import DataFormat, IngestionMappingKind
 from azure.kusto.data.exceptions import KustoClientError
 
 
-class KustoMappingAndMappingReferenceError(KustoClientError):
+class KustoMappingError(KustoClientError):
+    """
+    Raised when the provided mapping arguments are invalid.
+    """
+
+
+class KustoDuplicateMappingError(KustoMappingError):
     """
     Error to be raised when ingestion properties include both
-    a mapping and a mapping reference
+    column mappings and a mapping reference
     """
 
     def __init__(self):
-        message = "Ingestion properties contain both an explicit mapping and a mapping reference."
-        super(KustoMappingAndMappingReferenceError, self).__init__(message)
-
-
-class KustoDuplicateMappingError(KustoClientError):
-    """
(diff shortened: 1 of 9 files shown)


azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -1,39 +1,39 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
-
+from azure.kusto.data.data_format import DataFormat, IngestionMappingKind
 from azure.kusto.data.exceptions import KustoClientError
 
 
-class KustoMappingAndMappingReferenceError(KustoClientError):
+class KustoMappingError(KustoClientError):
```
yihezkel, line 7: What's the benefit of another layer in the hierarchy?
I'm not sure it adds value, especially when we're trying to move away from most mapping errors (including `KustoMissingMappingError`)

AsafMah: I needed an exception to throw that is related to Mapping, but didn't fit any of the ones we had so I created the general "KustoMappingError", and from there if I was a user, I'd expect that catching a "KustoMappingError" will catch every type of mapping error, so hence the inhertience.

yihezkel: But there are only 2. is it worth adding a hierarchy to group together 2 exceptions that don't share any properties, but instead share 1 common theme?
Also, 1 of those 2 will be disappearing soon.

Is it really surprising to a user to get raised a `KustoDuplicateMappingError(KustoClientError)` or a `KustoMissingMappingReferenceError(KustoClientError)`?

If we look at https://www.tutorialspoint.com/java/java_builtin_exceptions.htm we see many exceptions (e.g. `ArrayIndexOutOfBoundsException`, `ArrayStoreException`, `NegativeArraySizeException` all relating to arrays) that share a common theme, but since they don't share uniquely common members, they don't have an OO hierarchy.

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -1,39 +1,39 @@
...
-        message = "Ingestion properties contain both an explicit mapping and a mapping reference."
-        super(KustoMappingAndMappingReferenceError, self).__init__(message)
-
 
-class KustoDuplicateMappingError(KustoClientError):
+class KustoDuplicateMappingError(KustoMappingError):
     """
-    Error to be raised when ingestion properties include two explicit mappings.
+    Error to be raised when ingestion properties include both
+    column mappings and a mapping reference
```
yihezkel, line 16: nit: "Error to be raised when " is unnecessarily verbose

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
AsafMah: Removed hierarchy

azure-kusto-ingest/azure/kusto/ingest/exceptions.py
```diff
@@ -1,39 +1,39 @@
...
-    def __init__(self, data_format: str):
-        message = f"When stream format is {data_format}, mapping name must be provided."
-        super(KustoMissingMappingReferenceError, self).__init__(message)
+    @staticmethod
+    def required_data_format(data_format: DataFormat) -> "KustoMissingMappingError":
+        return KustoMissingMappingError(f"When stream format is '{data_format.kusto_value}', a mapping must be provided.")
+
+    @staticmethod
+    def required_with_kind(kind: IngestionMappingKind) -> "KustoMissingMappingError":
+        return KustoMissingMappingError(f"When ingestion mapping kind is set ('{kind.value}'), a mapping must be provided.")
```
yihezkel, line 36: nit: I'm surprised to see hardcoded static methods within the exception class that return the exception constructed with predefined text.
I'd either expect the caller to provide the text, or for the caller to provide the text based on a constant (like https://github.com/Azure/azure-kusto-java/blob/master/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java#L111).

I guess there are different approaches, so ignore if you prefer this approach. But if my experience is enough to go by (definitely might not be), this is a non-standard approach.

yihezkel (approved):
Sweet!

PR #367, opened 2022-01-16
E2E - Only clear cache once

AsafMah:
We only need to clear the cache after we create the table.
This PR removes superfluous calls.

### DIFF
--- a/azure-kusto-ingest/tests/e2e.py
+++ b/azure-kusto-ingest/tests/e2e.py
@@ -38,8 +38,6 @@ from azure.kusto.ingest import (
     ManagedStreamingIngestClient,
 )
 
-CLEAR_DB_CACHE = ".clear database cache streamingingestion schema"
-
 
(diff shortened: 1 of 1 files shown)


azure-kusto-ingest/tests/e2e.py
```diff
@@ -207,6 +205,8 @@ def setup_class(cls):
 
         cls.client.execute(cls.test_db, f".alter table {cls.test_table} policy streamingingestion enable ")
 
+        cls.client.execute(cls.test_db, ".clear database cache streamingingestion schema")
```
yihezkel, line 208: nit: comment

yihezkel (dismissed):

azure-kusto-ingest/tests/e2e.py
AsafMah: Done

yihezkel (approved):

PR #376, opened 2022-01-25
Save space by using numpy datatype "int32" instead of "int64" for ADX int

yihezkel:
#### Pull Request Description

Save space by using numpy datatype "int32" instead of "int64" for ADX datatype of "int" (which is a 32-bit integer).

---

#### Future Release Comment
**Fixes:**
- Save space by using numpy datatype "int32" instead of "int64" for ADX datatype of "int" (which is a 32-bit integer).


AsafMah (approved):
Looks good

Issue #378, opened 2022-02-25
Proxy Support in Ingest Client


mitsuo0114:
**Is your feature request related to a problem? Please describe.**
I'm frustrated when we want to send data to ADX from firewalled network.
`https_proxy` environment variable works, but because it is global settings of python process, we need to use proxy as parameter.

Recent change https://github.com/Azure/azure-kusto-python/pull/360 was affected only query side but in my understanding, still we cannot use proxy in IngestClient 

Therefore, my script stopped because it cannot reach blob url.

**Describe the solution you'd like**

In terms of implementation, I don't have good idea to solve. (because of the complexity of implementation)
but I want to use proxy settings without environment variable setting.
And I also find a clue to solve, hopefuly my Additional context in later is helpful

**Describe alternatives you've considered**
None

**Additional context**
1)
Here is my code snipets to upload data with `QueuedIngestClient:ingest_from_file`.

```python
from azure.kusto.data import KustoConnectionStringBuilder
from azure.kusto.ingest import QueuedIngestClient, IngestionProperties, FileDescriptor
from azure.kusto.data.data_format import DataFormat

def postdata(self, dictdata, database, table, json_ingestion_mapping):
        ingestion_props = IngestionProperties(database=database,
                                              table=table,
                                              data_format=DataFormat.MULTIJSON,
                                              ingestion_mapping_reference=json_ingestion_mapping)
        kcsb = KustoConnectionStringBuilder.with_aad_application_key_authentication(self.cluster_url,
                                                                                    self.client_id,
                                                                                    self.client_secret,
                                                                                    self.authority_id)
        client = QueuedIngestClient(kcsb)
        client._resource_manager._kusto_client.set_proxy(proxy_url=self.proxy)

        with tempfile.NamedTemporaryFile("w") as tp:
            tp.write(json.dumps(dictdata))
            tp.seek(0)
            file_descriptor = FileDescriptor(tp.name)
            client.ingest_from_file(file_descriptor, ingestion_properties=ingestion_props)
```

Thanks to https://github.com/Azure/azure-kusto-python/pull/360, we can set proxy to ksuto_client in resource_manager. 
(still need access non-public parameters though)
```python
client._resource_manager._kusto_client.set_proxy(proxy_url=self.proxy)
```

But, in my understanding, because following implementation creates blob client & upload data without using any injected config, we cannot injected proxy config.
[...]

PR #379, opened 2022-02-25
Add parameter for blob client init (#378)

mitsuo0114:
#### Pull Request Description

This is sample implementation for https://github.com/Azure/azure-kusto-python/issues/378
(I don't think this is best solution but just want to show my thought)


AsafMah:
For now that's not the approach we want to take, since it may lead to unexpected results (in this case - passing proxy as a parameter will not pass it to resource manager).

I am working on a proper solution for proxies for ingest that should be out soon.

PR #380, opened 2022-03-01
Add proxy support to ingest

AsafMah:
Adds a "set_proxy" method for all ingest clients.
Also fixes the names of some tests.

Tested manually since automatic tests of this would be very taxing.

solves #378

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
@@ -140,3 +140,6 @@ class _ResourceManager:
             return command_result.primary_results[0][0][_SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
             return ""
+
+    def set_proxy(self, proxy_url: str):
+        self._kusto_client.set_proxy(proxy_url)
--- a/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
@@ -78,6 +78,13 @@ class BaseIngestClient(metaclass=ABCMeta):
         """
         pass
 
+    @abstractmethod
+    def set_proxy(self, proxy_url: str):
+        """Set proxy for the ingestion client.
+        :param str proxy_url: proxy url.
+        """
+        pass
+
     def ingest_from_dataframe(self, df: "pandas.DataFrame", ingestion_properties: IngestionProperties) -> IngestionResult:
         """Enqueue an ingest command from local files.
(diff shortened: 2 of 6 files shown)


azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -140,3 +140,6 @@ def retrieve_service_type(self):
             return command_result.primary_results[0][0][_SERVICE_TYPE_COLUMN_NAME]
         except (TypeError, KeyError):
             return ""
+
+    def set_proxy(self, proxy_url: str):
```
ArielYehezkely, line 144: Better to just create a new ResourceManager

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
```diff
@@ -78,6 +78,13 @@ def ingest_from_stream(self, stream_descriptor: Union[StreamDescriptor, IO[AnySt
         """
         pass
 
+    @abstractmethod
+    def set_proxy(self, proxy_url: str):
```
ArielYehezkely, line 82: I think it's better to set it through optional ctor parameter

AsafMah: In the PR for the client you convinced me to not make it a ctor parameter - https://github.com/Azure/azure-kusto-python/pull/360/files/efde7dd3245bfec121eaf4b5958bbbdac3333557#diff-ed58329617a49a8000e9b4c91a78ad8be106646e8f41e37403d57e2dc48f09c5

What's different here?

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
AsafMah: The change is only in the kusto client, and since I don't think we want to create a new client (since we don't want to restart the token provider), creating a new resource manager with the same client wouldn't change anything.

ArielYehezkely: I'm mean create the ResourceManager with the modified client instead of exposing a way to set a property of a member

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
ArielYehezkely: The difference is the proxy is a setting of the KustoClient, and here you set a property of a member - if we didn't need to use the proxy for storage as well, I'd say we should add a way to create a client with KustoClient in addition to creation with a uri or kcsb

ArielYehezkely (approved):

PR #381, opened 2022-03-07
Add retry mechanism for potentially throttled processes.

alonadam:
#### Pull Request Description
Retry mechanism added to support retry of potentially throttled (by DM service) processes.
 It uses existing code for exponential retry.

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


azure-kusto-ingest/azure/kusto/ingest/_retry.py
```diff
@@ -11,8 +16,28 @@ def __init__(self, max_attempts, sleep_base_secs: float = 1.0, max_jitter_secs:
...
+        """
         self.current_attempt += 1
+        if self:
+            sleep((self.sleep_base_secs * (2**self.current_attempt)) + (random.uniform(0, self.max_jitter_secs)))
 
     def __bool__(self):
         return self.current_attempt < self.max_attempts
+
+
+def func_retry_wrapper(func, max_retry, sleep_base_secs: float, *exceptions):
```
AsafMah, line 30: The name should indicate that it's exponential.

AsafMah, line 30: Ideally I'd like it to be a decorator, but because it receives parameters that are dependent on class members that's hard.

Do you have an idea as to how or do you prefer it like this?

AsafMah, line 30: typing: `*exceptions: Tuple[Exception, ...]`

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -88,7 +93,9 @@ def _get_resource_by_name(self, table: KustoResultTable, resource_name: str):
         return [_ResourceUri.parse(row["StorageRoot"]) for row in table if row["ResourceTypeName"] == resource_name]
 
     def _get_ingest_client_resources_from_service(self):
```
AsafMah, line 95: Do we want throttling checks only on get resources? why not all queries? what sparked this PR?

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -88,7 +93,9 @@ def _get_resource_by_name(self, table: KustoResultTable, resource_name: str):
...
-        table = self._kusto_client.execute("NetDefaultDB", ".get ingestion resources").primary_results[0]
+        table = func_retry_wrapper(
+            partial(self._kusto_client.execute, "NetDefaultDB", ".get ingestion resources"), self._max_retry, self._seconds_between_retry, KustoThrottlingError
```
AsafMah, line 97: Personally, I think a small lambda is more readable than a partial, and either way I'd separate the first arguments from the rest for readability.

azure-kusto-ingest/azure/kusto/ingest/_retry.py
AsafMah, line 30: Can we generalize it to also be used in ManagedStreaming?

AsafMah (commented):
Some comments.


And obviously missing tests

azure-kusto-ingest/azure/kusto/ingest/_retry.py
alonadam: I was also thinking about decorator but than the "retry scope" is the entire decorated function (even though you may only want to retry a specific part or line). I feel like this is more practical. I agree that decorator is more appealing.

AsafMah: Correct me if I'm wrong, but this should be less of a problem since we only retry on the specific exception.
So even if we cover "too much" in the retry we won't accidentally retry something we don't

alonadam: You are correct but there might be more complex scenarios in the future and this approach is more flexible.
I don't mind adding a decorator and let the developer choose as he sees fit. Are you OK with having both options (decorator | direct call to 'func_retry_wrapper')?

AsafMah: The python zen is "There should be one-- and preferably only one --obvious way to do it.", I don't think we should put code in if we're not going to use it.

A decorator is just as flexible, you can call it like this if you want
```python
my_decorator(partial(self._kusto_client.execute, "NetDefaultDB", ".get ingestion resources"))()
```

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
alonadam: Only for get resources and get token. These two operations are the most likely to get throttled.

dev_requirements.txt
```diff
@@ -6,4 +6,5 @@ black;python_version >= '3.6'
 aioresponses>=0.6.2
 pytest-asyncio>=0.12.0
 azure-core>=1.11.0
-asgiref>=3.2.3
\ No newline at end of file
+asgiref>=3.2.3
+tenacity>=8.0.0
```
AsafMah, line 10: You don't want to add this here, this is dev requirements - meaning dependencies you only need when working on the repo.
You want to put it in azure-kusto-ingest's setup.py

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
```diff
@@ -75,6 +77,16 @@ def __init__(self, kusto_client: KustoClient):
         self._authorization_context = None
         self._authorization_context_last_update = None
 
+        self.__set_throttling_settings()
+
+    def __set_throttling_settings(self, num_of_attempts: int = 4, max_seconds_per_retry: float = 30):
```
AsafMah, line 82: Unit test this? (make sure it rethrows other exceptions and retries with KustoThrottlingError)

AsafMah (commented):
Almost done!

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
alonadam: In my opinion this is a bad practice - you should not write unit tests to verify code that is not yours.
In this case we use external library to implement retry policy, but what if we used external library to parse JSON file? Would I then needed to write unit tests to verify that parsing works as expected?

dev_requirements.txt
alonadam: ok fixed.

azure-kusto-ingest/azure/kusto/ingest/_resource_manager.py
AsafMah: Resolved offline

AsafMah (approved):

PR #382, opened 2022-03-09
Update cryptography requirement to 35.0.0 or above

AsafMah:
To comply with some security standards

### DIFF
--- a/azure-kusto-data/setup.py
+++ b/azure-kusto-data/setup.py
@@ -43,6 +43,6 @@ setup(
     namespace_packages=["azure"],
     keywords="kusto wrapper client library",
     packages=find_packages(exclude=["azure", "tests"]),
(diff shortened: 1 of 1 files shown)


saoc90:
@AsafMah why did you close this PR? I would be interessted in the updated `cryptography` dependency as well. I'm currently working on an integration for Home Assistant where I have a version conflinct with the older version ´cryptography-3.4.8´.

yihezkel:
@saoc90 
We closed this PR because I did an analysis of the Kusto Python SDK's dependency tree, and found that it allows for cryptography==35.0.0.

What package in this SDK is preventing you from using v35.0.0?

saoc90:
@yihezkel I just double checked the issue. And it seems that the pip error message was missleading. The error was due to a missing build dependency in the alpine docker container. You're right the kusto package allows versions >=35.0.0.

Thank you!

PR #396, opened 2022-04-24
Split client.py into multiple files

AsafMah:
#### Pull Request Description

Split client.py into multiple files for easier readability.

---

#### Future Release Comment]
**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Split client.py into multiple files for easier readability.


azure-kusto-data/azure/kusto/data/client_request_properties.py
```diff
@@ -0,0 +1,51 @@
+import json
+from typing import Any
+
+from .kcsb import _assert_value_is_valid
```
danield137, line 4: does this import make sense? seems odd importing a private function from `kcsb`.

danield137 (dismissed):
thank you! much better :)

azure-kusto-data/azure/kusto/data/client_request_properties.py
AsafMah: Fixed

PR #401, opened 2022-06-07
Added tests to check parsing correctness of proxy URLs

enmoed:
#### Pull Request Description

Added testing for parsing of proxy URLs

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/azure-kusto-data/azure/kusto/data/kcsb.py
+++ b/azure-kusto-data/azure/kusto/data/kcsb.py
@@ -128,12 +128,13 @@ class KustoConnectionStringBuilder:
         for kvp_string in connection_string.split(";"):
             key, _, value = kvp_string.partition("=")
             keyword = self.ValidKeywords.parse(key)
+            value = value.strip()
             if keyword.is_str_type():
-                self[keyword] = value.strip()[:-1] if value.strip()[-1] == "/" else value.strip()
+                self[keyword] = value[:-1] if value[-1] == "/" else value
             if keyword.is_bool_type():
-                if value.strip() in ["True", "true"]:
+                if value in ["True", "true"]:
                     self[keyword] = True
-                elif value.strip() in ["False", "false"]:
+                elif value in ["False", "false"]:
                     self[keyword] = False
                 else:
                     raise KeyError("Expected aad federated security to be bool. Recieved %s" % value)
--- a/azure-kusto-data/azure/kusto/data/kcsb.py
+++ b/azure-kusto-data/azure/kusto/data/kcsb.py
@@ -129,7 +129,7 @@ class KustoConnectionStringBuilder:
             key, _, value = kvp_string.partition("=")
             keyword = self.ValidKeywords.parse(key)
             if keyword.is_str_type():
-                self[keyword] = value.strip()
+                self[keyword] = value.strip()[:-1] if value.strip()[-1] == "/" else value.strip()
             if keyword.is_bool_type():
                 if value.strip() in ["True", "true"]:
                     self[keyword] = True
--- a/azure-kusto-data/azure/kusto/data/kcsb.py
+++ b/azure-kusto-data/azure/kusto/data/kcsb.py
@@ -130,7 +130,7 @@ class KustoConnectionStringBuilder:
             keyword = self.ValidKeywords.parse(key)
             value_stripped = value.strip()
             if keyword.is_str_type():
-                self[keyword] = value_stripped[:-1] if value_stripped[-1] == "/" else value_stripped
+                self[keyword] = value_stripped.rstrip("/")
             elif keyword.is_bool_type():
                 if value_stripped in ["True", "true"]:
                     self[keyword] = True
--- a/azure-kusto-data/azure/kusto/data/kcsb.py
+++ b/azure-kusto-data/azure/kusto/data/kcsb.py
@@ -128,13 +128,13 @@ class KustoConnectionStringBuilder:
         for kvp_string in connection_string.split(";"):
             key, _, value = kvp_string.partition("=")
             keyword = self.ValidKeywords.parse(key)
-            value = value.strip()
+            value_stripped = value.strip()
             if keyword.is_str_type():
-                self[keyword] = value[:-1] if value[-1] == "/" else value
-            if keyword.is_bool_type():
-                if value in ["True", "true"]:
+                self[keyword] = value_stripped[:-1] if value_stripped[-1] == "/" else value_stripped
+            elif keyword.is_bool_type():
+                if value_stripped in ["True", "true"]:
                     self[keyword] = True
-                elif value in ["False", "false"]:
+                elif value_stripped in ["False", "false"]:
                     self[keyword] = False
                 else:
                     raise KeyError("Expected aad federated security to be bool. Recieved %s" % value)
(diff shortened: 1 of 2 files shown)


azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -158,3 +158,40 @@ def test_proxy_token_providers(self, mock_get, proxy_kcsb):
...
+            "/some_resource_group/providers/microsoft.operationalinsights/workspaces/some_workspace",
+            "https://ade.loganalytics.io:4242/subscriptions/da45f7ac-97c0-4fff-ac66-3b6810eb4f65/resourcegroups"
+            "/some_resource_group/providers/microsoft.operationalinsights/workspaces/some_workspace/;fed=true": "https://ade.loganalytics.io:4242/subscriptions/da45f7ac-97c0-4fff-ac66-3b6810eb4f65"
+            "/resourcegroups/some_resource_group/providers/microsoft.operationalinsights/workspaces"
+            "/some_workspace",
+            "https://kusto.aria.microsoft.com": "https://kusto.aria.microsoft.com",
+            "https://kusto.aria.microsoft.com/": "https://kusto.aria.microsoft.com",
+            "https://kusto.aria.microsoft.com/;fed=true": "https://kusto.aria.microsoft.com",
+        }
+        for entry_key, entry_value in tests.items():
```
AsafMah, line 194: NIT: new line between the dictionary and the check

AsafMah, line 194: More meaningful names? "expected"/"actual"?

azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -158,3 +158,40 @@ def test_proxy_token_providers(self, mock_get, proxy_kcsb):
...
+            kcsb = KustoConnectionStringBuilder.with_aad_user_token_authentication(entry_key, "test")
+            client = KustoClient(kcsb)
+            assert (entry_value, client._kusto_cluster)
```
AsafMah, line 197: Sadly that's the wrong form of assert, this will always return true.

You want to do 
`assert entry_value == client._kusto_clutser`

And we already see some of the tests fail

azure-kusto-data/tests/test_kusto_client.py
```diff
@@ -158,3 +158,40 @@ def test_proxy_token_providers(self, mock_get, proxy_kcsb):
...
+            "https://ade.loganalytics.io:4242/subscriptions/da45f7ac-97c0-4fff-ac66-3b6810eb4f65/resourcegroups"
+            "/some_resource_group/providers/microsoft.operationalinsights/workspaces/some_workspace/;fed=true": "https://ade.loganalytics.io:4242/subscriptions/da45f7ac-97c0-4fff-ac66-3b6810eb4f65"
+            "/resourcegroups/some_resource_group/providers/microsoft.operationalinsights/workspaces"
+            "/some_workspace",
+            "https://kusto.aria.microsoft.com": "https://kusto.aria.microsoft.com",
+            "https://kusto.aria.microsoft.com/": "https://kusto.aria.microsoft.com",
+            "https://kusto.aria.microsoft.com/;fed=true": "https://kusto.aria.microsoft.com",
+        }
+        for entry_key, entry_value in tests.items():
+            kcsb = KustoConnectionStringBuilder.with_aad_user_token_authentication(entry_key, "test")
```
AsafMah, line 195: `kcsb = KustoConnectionStringBuilder(entry_key)`

It's shorter

azure-kusto-data/tests/test_kusto_client.py
AsafMah, line 197: I don't like the access to the private member, maybe we should add a getter/make it public. @ArielYehezkely WDYT?

AsafMah (changes requested):
Not bad for a first PR!

azure-kusto-data/tests/test_kusto_client.py
enmoed: thanks

enmoed: will do

azure-kusto-data/tests/test_kusto_client.py
enmoed: @AsafMah Not sure we need - I think we can access the same data source through kcsb.data_source. WDYT?

azure-kusto-data/azure/kusto/data/kcsb.py
```diff
@@ -129,7 +129,7 @@ def __init__(self, connection_string: str):
             key, _, value = kvp_string.partition("=")
             keyword = self.ValidKeywords.parse(key)
             if keyword.is_str_type():
-                self[keyword] = value.strip()
+                self[keyword] = value.strip()[:-1] if value.strip()[-1] == "/" else value.strip()
```
ArielYehezkely, line 132: This way you're calling strip() twice, better to call once and then remove the slash

azure-kusto-data/azure/kusto/data/kcsb.py
```diff
@@ -128,12 +128,13 @@ def __init__(self, connection_string: str):
         for kvp_string in connection_string.split(";"):
             key, _, value = kvp_string.partition("=")
             keyword = self.ValidKeywords.parse(key)
+            value_stripped = value.strip()
             if keyword.is_str_type():
-                self[keyword] = value.strip()
-            if keyword.is_bool_type():
-                if value.strip() in ["True", "true"]:
+                self[keyword] = value_stripped[:-1] if value_stripped[-1] == "/" else value_stripped
```
yogilad, line 133: This sort of voodoo blood magic requires a comment about what's the expected structure of the string being parsed.
;)

yogilad: Is this trimming a trailing slash?
if so, you can use .rstrip("/")

azure-kusto-data/azure/kusto/data/kcsb.py
```diff
@@ -128,12 +128,13 @@ def __init__(self, connection_string: str):
         for kvp_string in connection_string.split(";"):
             key, _, value = kvp_string.partition("=")
             keyword = self.ValidKeywords.parse(key)
+            value_stripped = value.strip()
             if keyword.is_str_type():
-                self[keyword] = value.strip()
```
AsafMah, line 132: You probably still want to strip, in addition to removing the trailing slash.

azure-kusto-data/azure/kusto/data/kcsb.py
```diff
@@ -128,12 +128,13 @@ def __init__(self, connection_string: str):
...
-            if keyword.is_bool_type():
-                if value.strip() in ["True", "true"]:
+                self[keyword] = value_stripped.rstrip("/")
+            elif keyword.is_bool_type():
+                if value_stripped in ["True", "true"]:
```
AsafMah, line 135: MMMM, this means that "True/" is a valid value here, which is not.
You only need the regular strip() here for spaces and such, not removing the trailing slash.

azure-kusto-data/azure/kusto/data/kcsb.py
yihezkel: See line 131

AsafMah (approved):
Good job!

PR #403, opened 2022-06-28
t-ronmoneta/python sample app

ronmonetaMicro:
Only deviation i made from C# was moving the AuthenticationModeOptions enum from sample_app to Utils - due to python limitations.


quick_start/kusto_sample_config.json
```diff
@@ -1,30 +1,33 @@
...
-	]
+  "kustoUri": "https://help.kusto.windows.net",
+  "ingestUri": "https://ingest-help.kusto.windows.net",
+  "databaseName": "MyDatabase",
+  "tableName": "SampleTable",
+  "useExistingTable": true,
+  "alterTable": true,
+  "queryData": true,
+  "ingestData": true,
+  "AuthenticationMode": "UserPrompt",
```
AsafMah, line 10: Some properties are PascalCase and some are camelCase, was this also that way in c#?
That's a bit weird.

quick_start/sample_app.py
```diff
@@ -0,0 +1,382 @@
+import enum
+import json
+import uuid
+from typing import ClassVar
+import inflection as inflection
```
AsafMah, line 5: This package (and tqdm) doesn't come with python by default, so users will get an error.
I think the simplest thing to do here is to include a requirements.txt file and ask for users to install it

quick_start/sample_app.py
```diff
@@ -0,0 +1,382 @@
...
+from quick_start.utils import Utils, AuthenticationModeOptions
+from azure.kusto.data import DataFormat, KustoClient
+
+
+class SourceType(enum.Enum):
+    """
+    SourceType - represents the type of files used for ingestion
+    """
+
+    local_file_source = ("localFileSource",)
```
AsafMah, line 16: Why are the values tuples instead of just a string?

quick_start/sample_app.py
```diff
@@ -0,0 +1,382 @@
...
+    blob_source = ("blobSource",)
+
+
+class ConfigData:
+    """
+    ConfigData object - represents a file from which to ingest
+    """
+
+    source_type: ClassVar[SourceType]
```
AsafMah, line 25: Classvar in python roughly means the same as static variables in c#. Since these are all member variables, you don't need ClassVar (and in configjson too)

quick_start/sample_app.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        Converts dict object  - JSON style camelCase to ConfigData object - python snake_case
+        :param data_dicts: Dict of data sources list to ingest from - styled in camelCase
+        :return: ConfigData Data sources list to ingest from - styled in snake_case
+        """
+        config_data_list = []
+
+        for data_dict in data_dicts:
+            config_data = ConfigData()
+            config_data.source_type = SourceType[inflection.underscore(data_dict["sourceType"])]
```
AsafMah, line 124: If you are using inflection here, why not use it for all of the values? it would save you a lot of code

AsafMah (commented):
Overall looks good, some notes

quick_start/kusto_sample_config.json
ronmonetaMicro: True, it was due moving some of the class vars to the JSON. the same goes for C# and node. ill change it all to camelCase.

quick_start/sample_app.py
AsafMah: You use the infliction library to convert snake_case to camelCase for source_type.

But for all of the keys in data_dict, you manually convert them to snake_case, I'm saying you can use the library the save you this manual conversion.

quick_start/sample_app.py
```diff
@@ -0,0 +1,350 @@
...
+        try:
+            with open(config_file_name, "r") as config_file:
+                json_dict = json.load(config_file)
+
+        except Exception as ex:
+            Utils.error_handler(f"Couldn't read load config file from file '{config_file_name}'", ex)
+
+        json_dict["dataToIngest"] = cls.convert_config_data_fields(json_dict["data"])
+        json_dict.pop("data", None)
+        cls.config = ConfigJson(**{inflection.underscore(key): val for (key, val) in json_dict.items()})
```
AsafMah, line 79: Let's extract it into a function:
```python
    def keys_to_snake_case(json_dict):
        return {inflection.underscore(key): val for (key, val) in json_dict.items()}
```

And then it's more readable:
`ConfigJson(**keys_to_snake_case(json_dict))`

quick_start/sample_app.py
```diff
@@ -0,0 +1,350 @@
...
+        :param config_file_name: Configuration file path.
+        """
+        try:
+            with open(config_file_name, "r") as config_file:
+                json_dict = json.load(config_file)
+
+        except Exception as ex:
+            Utils.error_handler(f"Couldn't read load config file from file '{config_file_name}'", ex)
+
+        json_dict["dataToIngest"] = cls.convert_config_data_fields(json_dict["data"])
```
AsafMah, line 77: Turns out this can be a oneliner:
`json_dict["dataToIngest"] = cls.convert_config_data_fields(json_dict.pop("data"))`

quick_start/sample_app.py
```diff
@@ -0,0 +1,350 @@
...
+    def convert_config_data_fields(cls, data_dicts: list[dict]) -> list[ConfigData]:
+        """
+        Converts dict object  - JSON style camelCase to ConfigData object - python snake_case
+        :param data_dicts: Dict of data sources list to ingest from - styled in camelCase
+        :return: ConfigData Data sources list to ingest from - styled in snake_case
+        """
+        config_data_list = []
+
+        for data_dict in data_dicts:
+            config_data = ConfigData()
```
AsafMah, line 91: I know what we discussed in the chat, but don't you think this is nicer?
```python
            data_dict["sourceType"] = SourceType[inflection.underscore(data_dict.pop("sourceType"))]
            data_dict["dataFormat"] = DataFormat[data_dict.pop("format")]
            config_data_list.append(ConfigData(**KustoSampleApp.keys_to_snake_case(data_dict)))
```

If not, you can keep this as is.

quick_start/requirements.txt
```diff
@@ -0,0 +1,3 @@
+azure_storage==0.37.0
```
AsafMah, line 1: You need to have "azure.kusto.data" and "azure.kusto.ingest" in the requirements, and remove storage

quick_start/sample_app.py
```diff
@@ -0,0 +1,350 @@
+import enum
+import json
+import uuid
+from dataclasses import dataclass
+import inflection as inflection
+from azure.kusto.ingest import QueuedIngestClient
+from quick_start.utils import Utils, AuthenticationModeOptions
```
AsafMah, line 7: This requires the package to be installed, replace with:
`from utils import Utils, AuthenticationModeOptions`

quick_start/sample_app.py
```diff
@@ -0,0 +1,350 @@
...
+class SourceType(enum.Enum):
+    """
+    SourceType - represents the type of files used for ingestion
+    """
+
+    local_file_source = "localFileSource"
+    blob_source = "blobSource"
+
+
+class ConfigData:
```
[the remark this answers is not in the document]
ronmonetaMicro: will require to replace the code in convert_config_data_fields to inside the ConfigData parenthesis - which i think would look pretty bad (?). WDYT?

quick_start/README.md
```diff
@@ -30,7 +30,7 @@ You can use it as a baseline to write your own first kusto client application, a
 
 1. Open a command line window and navigate to the folder where you extracted the script.
 1. Run `python -m pip install azure-kusto-data azure-kusto-ingest`. If upgrading, append ` --upgrade`.
```
AsafMah, line 32: Change to:
`python -m pip install -r requirement.txt`

quick_start/requirements.txt
```diff
@@ -0,0 +1,4 @@
+azure-kusto-data==3.0.1
+azure-kusto-ingest==3.0.1
+inflection==0.5.1
+tqdm==4.64.0
```
AsafMah, line 4: For our packages, you can set it to non breaking versions:
For others, you can set it to minor versions up:
```
azure-kusto-data>=3.0.1,<4.0.0
azure-kusto-ingest>=3.0.1,<4.0.0

inflection~=0.5.1
tqdm~=4.64.0
```

quick_start/sample_app.py
```diff
@@ -0,0 +1,355 @@
...
+    @classmethod
+    def alter_merge_existing_table_to_provided_schema(cls, kusto_client: KustoClient, database_name: str, table_name: str, table_schema: str) -> None:
+        """
+        Alter-merges the given existing table to provided schema.
+        :param kusto_client: Client to run commands
+        :param database_name: DB name
+        :param table_name: Table name
+        :param table_schema: Table Schema
+        """
+        command = f".alter-merge table {table_name} {table_schema}"
```
AsafMah, line 152: There was a discussion about quoting the names we insert into queries, since it broke someone's flow with a table with a hyphen in it.
Did you talk to Yochai? should we do this here or in another PR?

AsafMah (commented):
Great Job!

quick_start/sample_app.py
ronmonetaMicro: Yep, he wanted me to fix it in all SA's, but in a different PR for each after we close everything else

AsafMah (approved):

PR #407, opened 2022-08-08
User/t eitanmoed/add-e2e-tests-kusto-data

enmoed:
#### Pull Request Description

Split e2e test suite into ingest and data suites. Added an e2e test for Application Insights flow.

---

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


azure-kusto-ingest/tests/e2e.py
```diff
@@ -58,9 +54,6 @@ class TestE2E:
     client: ClassVar[KustoClient]
     test_table: ClassVar[str]
     current_count: ClassVar[int]
-    ingest_client: ClassVar[QueuedIngestClient]
```
AsafMah, line 61: Why did you remove these?

azure-kusto-data/tests/e2e.py
```diff
@@ -0,0 +1,278 @@
...
+from azure.kusto.data.aio import KustoClient as AsyncKustoClient
+from azure.kusto.data.streaming_response import FrameType
+
+
+@pytest.fixture(params=["ManagedStreaming", "NormalClient"])
+def is_managed_streaming(request):
+    return request.param == "ManagedStreaming"
+
+
+class TestE2E:
```
AsafMah, line 24: I think the cloud info tests at the end of the ingest e2e can also go here

.github/workflows/pythonpackage.yml
```diff
@@ -44,8 +44,10 @@ jobs:
           TEST_DATABASE: ${{ secrets.TEST_DATABASE }}
           TEST_BLOB: ${{ secrets.TEST_BLOB }}
           ENGINE_CONNECTION_STRING: ${{ secrets.ENGINE_CONNECTION_STRING }}
+          APPLICATION_INSIGHTS_ENGINE_CONNECTION_STRING: ${{https://ade.applicationinsights.io/subscriptions/12534eb3-8109-4d84-83ad-576c0d5e1d06/resourcegroups/clients_e2e_test/providers/microsoft.insights/components/kusto-e2e-app-insights}}
```
AsafMah, line 47: The ${{}} syntax is for variables from github.
Since we hardcode them here, you don't need the ${{}}, just write the values. (That's why your tests don't run automatically)

AsafMah (changes requested):
Good job!

azure-kusto-ingest/tests/e2e.py
enmoed: mistake - will fix

azure-kusto-data/tests/e2e.py
enmoed: sounds good

.github/workflows/pythonpackage.yml
enmoed: fixed

azure-kusto-ingest/tests/e2e.py
enmoed: On second thought, any reason we initialize ingest_client + streaming_ingest_client + managed_streaming_ingest and not the ..._file_path variables declared in the setup_up func? In both instances these variables are only used in instance methods and not in class methods.. Whereas all other class initialized variables are used in class methods and must be declared

AsafMah: I don't understand, we currently initialize both of them in the setup function for the class.
We don't have a setup function for an instance, though that's another topic.
Either way, we would want to declare them, if they were instance variables we would just do it without the ClassVar, the reason is for better autocompletion and error reporting.

AsafMah (approved):

PR #412, opened 2022-08-24
remove unnecessary raise

pengzhengyi:
#### Pull Request Description

`_handle_http_error` will definitely raise exception, I believe another raise is unnecessary.

---

#### Future Release Comment
Remove raise in invocations of `_handle_http_error`.

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


yogilad:
Thanks for this contribution.
The point of the raise in these lines it to tell the IDE the function is not expected to continue after this line.
The called API will handle actual raise itself.

PR #413, opened 2022-09-01
ignorefirstrecord fix

ronmonetaMicro:
#### Pull Request Description

_[Add a description of your pull request here]_

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
@@ -166,6 +166,7 @@ class IngestionProperties:
         drop_by_tags: Optional[List[str]] = None,
         additional_tags: Optional[List[str]] = None,
         flush_immediately: bool = False,
+        ignore_first_record: bool = False,
         report_level: ReportLevel = ReportLevel.DoNotReport,
         report_method: ReportMethod = ReportMethod.Queue,
         validation_policy: Optional[ValidationPolicy] = None,
@@ -211,6 +212,7 @@ class IngestionProperties:
         self.ingest_by_tags = ingest_by_tags
         self.drop_by_tags = drop_by_tags
         self.flush_immediately = flush_immediately
+        self.ignore_first_record = ignore_first_record
         self.report_level = report_level
         self.report_method = report_method
         self.validation_policy = validation_policy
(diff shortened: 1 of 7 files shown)


azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
```diff
@@ -211,6 +212,7 @@ def __init__(
         self.ingest_by_tags = ingest_by_tags
         self.drop_by_tags = drop_by_tags
         self.flush_immediately = flush_immediately
+        self.ignore_first_record = ignore_first_record
```
yogilad, line 215: The name of the ingestion properties json rep for this value is 'ignoreFirstRecord'
What makes this be written correctly when sending the request.

Can we add an E2E test to make sure it works as expected?
there are already E2Es ingesting CSVs so it should be simple

quick_start/utils.py
```diff
@@ -207,9 +218,10 @@ def ingest_from_file(
             :param table_name: Table name
             :param file_path: File path
             :param data_format: Given data format
+            :param ignore_first_record: Flag noting whether to ignore the first record in the table
```
yogilad, line 221: It should be made clear this is relevant for tabular textual formats (CSV and the likes).
You can also point to the below documentation
https://docs.microsoft.com/en-us/azure/data-explorer/ingestion-properties

azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
ronmonetaMicro: Not sure i understood correctly - this attribute will receive data from outside, not directly from the JSON - so the naming here doesnt really matter (as long as the JSON is being read correctly somewhere else). did you mean something different?

AsafMah: Hey yochai, we are using the library inflection to convert the case from camelcase to snakecase - https://github.com/Azure/azure-kusto-python/blob/ronmoneta/ignorefirestrecord/quick_start/sample_app.py#L39

We don't need e2e tests here I think, or if we did - this change doesn't warrants them.

I'm resolving

AsafMah (approved):

PR #418, opened 2022-09-12
Added closability to the clients

AsafMah:
Adds the option the close all query and ingest clients.


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -72,6 +74,17 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
         self._session.mount("http://", adapter)
         self._session.mount("https://", adapter)
 
+    def close(self):
+        if not self._is_closed:
+            self._session.close()
+        super().close()
+
+    def __enter__(self):
```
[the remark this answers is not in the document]
yogilad: Maybe move it to the base?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -50,6 +55,8 @@ def ingest_from_file(self, file_descriptor: Union[FileDescriptor, str], ingestio
         :param file_descriptor: a FileDescriptor to be ingested.
         :param azure.kusto.ingest.IngestionProperties ingestion_properties: Ingestion properties.
         """
+        super().ingest_from_file(file_descriptor, ingestion_properties)
```
yogilad, line 58: Is this a bug fix?
What happened before?
(Same for below stream API)

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -91,19 +100,22 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
             raise ex
 
         random_queue = random.choice(queues)
-        queue_service = QueueServiceClient(random_queue.account_uri, proxies=self._proxy_dict)
-        authorization_context = self._resource_manager.get_authorization_context()
-        ingestion_blob_info = IngestionBlobInfo(blob_descriptor, ingestion_properties=ingestion_properties, auth_context=authorization_context)
-        ingestion_blob_info_json = ingestion_blob_info.to_json()
-        queue_client = queue_service.get_queue_client(queue=random_queue.object_name, message_encode_policy=TextBase64EncodePolicy())
-        queue_client.send_message(content=ingestion_blob_info_json, timeout=self._SERVICE_CLIENT_TIMEOUT_SECONDS)
+        with QueueServiceClient(random_queue.account_uri, proxies=self._proxy_dict) as queue_service:
```
yogilad, line 113: Should we not reuse those?

azure-kusto-data/azure/kusto/data/client.py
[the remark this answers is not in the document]
AsafMah: `__enter__` is used for "with",` __aenter__` is used for async with 
Since it's wrong to initialize an async client with a normal "with", the method is not available

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
AsafMah: Not a bug fix. Before this commit, the super() method was fully abstract. Now it checks if the ingestor is closed, so we need to go to base.

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
AsafMah: I can make a cache for them, but just to be clear - they weren't re-used before, they were only not closed.

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -30,8 +30,13 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
...
         return self
 
-    def __aexit__(self, exc_type, exc_val, exc_tb):
-        return self._session.__aexit__(exc_type, exc_val, exc_tb)
+    async def close(self):
+        if not self._is_closed:
+            await self._session.__aexit__(None, None, None)
+        super().close()
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
```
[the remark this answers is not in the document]
AsafMah: Hmmm
We can maybe give them `__enter__` and `__exit__`, but it's useless since they are not exposed to the user, and are only ever used as memebers (as opposed to local variables).
`__aenter__` doesn't make sense as they don't have an async initialization

azure-kusto-data/README.rst
```diff
@@ -14,14 +14,13 @@ Overview
     authority_id = "<insert here your AAD tenant id>"
 
     kcsb = KustoConnectionStringBuilder.with_aad_application_key_authentication(cluster, client_id, client_secret, authority_id)
-    client = KustoClient(kcsb)
+    with KustoClient(kcsb) as client:
```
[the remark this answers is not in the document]
AsafMah: The sample or the PR?
We can add a comment explaining, but it is good practice to close the client after you're fully done with it

azure-kusto-data/azure/kusto/data/aio/client.py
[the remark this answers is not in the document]
yogilad: Lost you there. Is it something having to do with async python?
f2f?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: So every time we enqueue a message we create a new blob client?
Seems like something that needs to be encapsulated by the resource manager, no?
Separate PR for sure though

AsafMah: I'll open an issue

yogilad (approved):

PR #432, opened 2022-11-17
Pass settings to http parent

AsafMah:
We popped the settings, so they were never passed to the parent. This PR fixes it.


alonadam (approved):

AsafMah:
Superseeded by the other PR

PR #459, opened 2023-02-21
Close resources improvements

AsafMah:
* TokenProviders are now context managers. The tests properly close them
* Added close to tests that are missing
* Properly close token provider tokens

### DIFF
--- a/azure-kusto-data/azure/kusto/data/kusto_trusted_endpoints.py
+++ b/azure-kusto-data/azure/kusto/data/kusto_trusted_endpoints.py
@@ -1,13 +1,11 @@
 import copy
-import itertools
 import json
-import os
-from typing import List, Iterable, Dict
+from pathlib import Path
+from typing import List, Dict
 from urllib.parse import urlparse
 
 from azure.kusto.data.helpers import get_string_tail_lower_case
 from azure.kusto.data.security import _is_local_address
-
 from .exceptions import KustoClientInvalidConnectionStringException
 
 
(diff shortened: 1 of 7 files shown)


azure-kusto-data/azure/kusto/data/kusto_trusted_endpoints.py
```diff
@@ -103,6 +101,7 @@ def set_override_policy(self, matcher):
         self._override_matcher = matcher
 
 
-_filename = os.path.join(os.path.split(os.path.abspath(__file__))[0], "wellKnownKustoEndpoints.json")
-_well_known_kusto_endpoints_data = json.load(open(_filename))
+_filename = Path(__file__).absolute().parent / "wellKnownKustoEndpoints.json"
```
yihezkel, line 104: I'm unfamiliar with this syntax, but this seems off to me, and I'm also concerned about platform independence. Can you use a `join()` function instead?

AsafMah: It's identical to join - https://docs.python.org/3.7/library/pathlib.html#operators

I can use the joinpath() method if you prefer though.

yihezkel: If anyone else is like me, I do think `joinpath()` would be clearer.

Approving (with this suggestion)

yihezkel (dismissed):

yihezkel (approved):

PR #460, opened 2023-02-21
Use bytes instead of string in tests

AsafMah:
Use bytes instead of strings in tests:
1. More correct - that's how we get the data
2. Solves warnings

### DIFF
--- a/azure-kusto-data/tests/test_streaming_query.py
+++ b/azure-kusto-data/tests/test_streaming_query.py
@@ -1,5 +1,5 @@
 import os
-from io import StringIO
+from io import BytesIO
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


azure-kusto-data/tests/test_streaming_query.py
```diff
@@ -213,10 +213,10 @@ async def test_exception_in_row_async(self):
 
 class TestJsonTokenReader:
     def get_reader(self, data) -> JsonTokenReader:
-        return JsonTokenReader(StringIO(data))
+        return JsonTokenReader(BytesIO(data.encode("utf-8")))
```
yihezkel, line 216: nit: I think UTF-8 is the default anyway

yihezkel (approved):

azure-kusto-data/tests/test_streaming_query.py
AsafMah: Thanks

PR #464, opened 2023-03-07
0 redirects by default

AsafMah:


azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str]):
         super().__init__(kcsb, True)
 
         self._session = ClientSession()
+        self._max_redirects = 0
```
yogilad, line 34: @AsafMah , is this a mistake?
Seems this should have been,
```suggestion
        self._session.max_redirects = 0
```

AsafMah: This is not, in our async client you can't change the redirects at the session level, only at the request level.
Further down the file, we use _max_redirects for that.

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -94,6 +95,9 @@ def set_proxy(self, proxy_url: str):
         super().set_proxy(proxy_url)
         self._session.proxies = {"http": proxy_url, "https": proxy_url}
 
+    def set_redirect_count(self, redirect_count: int):
+        self._session.max_redirects = redirect_count
```
yogilad, line 99: Are we missing tracing changes here?

azure-kusto-data/tests/e2e.py
```diff
@@ -300,3 +303,33 @@ def test_cloud_info(self):
     def test_cloud_info_404(self):
         cloud_info = CloudSettings.get_cloud_info_for_cluster("https://www.microsoft.com")
         assert cloud_info is CloudSettings.DEFAULT_CLOUD
+
+    def test_redirects(self):
+        redirect_codes = [301, 302, 303, 307, 308]
+        with KustoClient("https://httpstat.us/") as client:
```
yogilad, line 309: How does this pass the DNS checks?
Should we not set an allow policy somewhere for this URI?

azure-kusto-data/tests/e2e.py
```diff
@@ -300,3 +303,33 @@ def test_cloud_info(self):
...
+            for code in redirect_codes:
+                client._query_endpoint = f"https://httpstat.us/{code}"
+                with pytest.raises(TooManyRedirects):
+                    client.execute("db", "table")
+            client.set_redirect_count(5)
```
yogilad, line 314: nit: a new line after scope back would be nice :)

azure-kusto-data/azure/kusto/data/_cloud_settings.py
```diff
@@ -87,9 +87,9 @@ def get_cloud_info_for_cluster(cls, kusto_uri: str, proxies: Optional[Dict[str,
             url = urljoin(kusto_uri, METADATA_ENDPOINT)
 
             # trace http get call for result
-            http_trace_attributes = KustoTracingAttributes.create_http_attributes(url=url, method="GET")
+            http_trace_attributes = KustoTracingAttributes.create_http_attributes(url=url, method="GET", max_redirects=0)
```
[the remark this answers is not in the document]
AsafMah: We do not (currently) pass the client to cloudsettings. doing this would be a big change in architecture here. I think it's out of scope for this PR, unless you want to postpone it until possible

azure-kusto-data/tests/e2e.py
AsafMah: The logic currently validates by cloud, since we use no token provider, we don't fetch cloud data, so we can't check the endpoints. I don't know if it's good or bad, but that's the logic.

azure-kusto-data/tests/e2e.py
AsafMah: Fixed

azure-kusto-data/azure/kusto/data/_telemetry.py
```diff
@@ -78,7 +73,7 @@ def create_streaming_ingest_attributes(cls, cluster: str, database: str, table:
         return ingest_attributes
 
     @classmethod
-    def create_http_attributes(cls, method: str, url: str, headers: dict = None) -> dict:
+    def create_http_attributes(cls, method: str, url: str, max_redirects: int, headers: dict = None) -> dict:
```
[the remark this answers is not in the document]
AsafMah: Fixed

PR #475, opened 2023-04-30
Added device code callback param

AsafMah:
### Added
- Added callback parameter to device code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+## Unreleased
+### Added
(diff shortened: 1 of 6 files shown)


yogilad:
In retrospect the below code was not so good.
The callback interface should actually have been callback(msg:string, uri:string, code:string)
Mind it does not even pass the dictionary, but only the message

``` python
 def _get_token_impl(self) -> Optional[dict]:
        flow = self._msal_client.initiate_device_flow(scopes=self._scopes)
        try:
            if self._device_code_callback:
                self._device_code_callback(flow[TokenConstants.MSAL_DEVICE_MSG])
            else:
                print(flow[TokenConstants.MSAL_DEVICE_MSG])
```

yogilad (approved):

caseyzhang123:
I'm super excited to use this new feature!  Do you know approximately when it might be released into a new version of the azure.kusto.data pip package?

AsafMah:
Today - 4.2.0 is out

PR #477, opened 2023-05-03
Default DB + parsing

AsafMah:
### Added
- Added Initial Catalog (Default Database) parameter to ConnectionStringBuilder
- Added method to manually set the cache for CloudSettings
### Changed
- Urls with one item after the path (i.e https://test.com/abc) will now be treated as cluster and initial catalog (ie. the cluster is "https://test.com" and the initial catalog is "abc").
    * This is to align our behaviour with the .NET SDK
### Fixed
- Some edge cases in url parsing

### DIFF
--- a/azure-kusto-data/azure/kusto/data/client_base.py
+++ b/azure-kusto-data/azure/kusto/data/client_base.py
@@ -45,6 +45,9 @@ class _KustoClientBase(abc.ABC):
         # notice that in this context, federated actually just stands for aad auth, not aad federated auth (legacy code)
         self._aad_helper = _AadHelper(self._kcsb, is_async) if self._kcsb.aad_federated_security else None
 
+        if not self._kusto_cluster.endswith("/"):
+            self._kusto_cluster += "/"
+
         # Create a session object for connection pooling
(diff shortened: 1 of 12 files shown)


azure-kusto-data/azure/kusto/data/client_base.py
```diff
@@ -58,6 +61,11 @@ def __init__(self, kcsb: Union[KustoConnectionStringBuilder, str], is_async):
         self.client_details = self._kcsb.client_details
         self._is_closed: bool = False
 
+        self.default_database = self._kcsb.initial_catalog
+
+    def get_database(self, database_name: str) -> str:
```
yogilad, line 66: This can be a property as in KCSB

yogilad (approved):
Approved with some suggestions

azure-kusto-data/azure/kusto/data/client_base.py
AsafMah: There is a "inital_catalog" property in kcsb, that's only a small helper for client
It should probably be private too

PR #481, opened 2023-06-14
Sanitize secrets from blob URL by removing all query parameters

yihezkel:

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
@@ -59,7 +59,10 @@ class IngestionResult:
         self.blob_uri = blob_uri
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
```diff
@@ -59,7 +59,10 @@ def __init__(self, status: IngestionStatus, database: str, table: str, source_id
         self.blob_uri = blob_uri
 
     def __repr__(self):
-        return f"IngestionResult(status={self.status}, database={self.database}, table={self.table}, source_id={self.source_id}, blob_uri={self.blob_uri})"
+        # Remove query parameters from blob_uri, if exists
+        if isinstance(self.blob_uri, str):
```
yihezkel, line 63: Else, should be no difference in functionality, since we previously traced `self.blob_uri`

AsafMah (approved):

PR #482, opened 2023-07-09
Sanitize secrets from blob URL by removing semicolon parameters

yihezkel:
#### Pull Request Description

Sanitize secrets from blob URL by removing semicolon parameters (previously only removed query parameters)

---

#### Future Release Comment
**Features:**
- Sanitize secrets from blob URL by removing semicolon parameters (previously only removed query parameters)


AsafMah (approved):
looks good

PR #487, opened 2023-07-12
Fixed default cloud test

AsafMah:
### Fixed
* Fixed test not working in dev envs
* Fixed build w/ black

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_cloud_settings.py
+++ b/azure-kusto-data/azure/kusto/data/_cloud_settings.py
@@ -1,3 +1,4 @@
+import dataclasses
 import os
 from threading import Lock
 from typing import Optional, Dict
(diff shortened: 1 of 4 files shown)


azure-kusto-data/azure/kusto/data/_cloud_settings.py
```diff
@@ -18,43 +19,24 @@
 DEFAULT_PUBLIC_LOGIN_URL = "https://login.microsoftonline.com"
 DEFAULT_REDIRECT_URI = "https://microsoft/kustoclient"
 DEFAULT_KUSTO_SERVICE_RESOURCE_ID = "https://kusto.kusto.windows.net"
+DEFAULT_DEV_KUSTO_SERVICE_RESOURCE_ID = "https://kusto.dev.kusto.windows.net"
```
yihezkel, line 22: Can we not get these from the Metadata endpoint?

yihezkel (approved):
See comment, please. Approved.

azure-kusto-data/azure/kusto/data/_cloud_settings.py
AsafMah: The whole point of these is to have a default when we are unable to get it from metadata (originally it was for clusters that didn't deploy the feature yet)

AsafMah: Resolved offline

PR #501, opened 2023-10-09
Test Python 3.11 and add to classifiers

tonybaloney:

PR #502, opened 2023-10-09
Test and add support for Python 3.12

tonybaloney:
this PR is a branch of #501 (that needs merging first)


AsafMah:
Thank you again!
I think we will wait with this one until aiohttp 3.9 is out of beta.

PR #505, opened 2023-10-30
Stop reliying on external tables + improve doc

AsafMah:
- Updated Docs
- Streaming test now creates its own table, and tests on a bigger area
- renamed test files to fit the convension


CONTRIBUTING.md
```diff
@@ -8,13 +8,41 @@ In order to run E2E tests, you need a Kusto database you have admin rights on.
...
 ```
 
+The E2E tests authenticate with DefaultAzureCredential, and will fall back to interactive login if needed.
+
+
+For more information on DefaultAzureCredential, see:  
+https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python
+
+To run the optional `token_provider` tests, you will need to set the booleans at the top of the file `test_token_providers.py` and the following environment variables in addition to the previous ones:
+```shell
```
ohadbitt, line 28: Why you put this as shell code ? this is not something you can run as is

azure-kusto-ingest/tests/test_e2e_ingest.py
```diff
@@ -152,19 +152,18 @@ def get_file_path() -> str:
...
+            else AsyncDefaultAzureCredential(exclude_interactive_browser_credential=False),
+        )
 
     @classmethod
     def dm_kcsb_from_env(cls) -> KustoConnectionStringBuilder:
-        if all([cls.app_id, cls.app_key, cls.auth_id]):
-            return KustoConnectionStringBuilder.with_azure_token_credential(cls.dm_cs, credential=DefaultAzureCredential())
-        else:
-            return KustoConnectionStringBuilder.with_interactive_login(cls.dm_cs)
+        return KustoConnectionStringBuilder.with_azure_token_credential(
```
ohadbitt, line 164: No need async?

ohadbitt (approved):

CONTRIBUTING.md
AsafMah: 1.you can in bash
2. it's nice syntax highlighting

azure-kusto-ingest/tests/test_e2e_ingest.py
AsafMah: We don't have async support for ingestion (yet..)

Issue #417, opened 2022-09-07
Unclosed client session


prashant-nigam:
#### Code Sample

```python
async with KustoClient(kcsb) as client:
    response = await client.execute(dbname, query,properties=clientReqProperties)


```
#### Problem description

I am initializing client using 'with' but still getting below errors:

[Error] Unclosed client sessionclient_session: <aiohttp.client.ClientSession object at 0x7ffa30df8310>
2022-09-07T10:10:42.288 
[Error] Unclosed connectorconnections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7ffa30e136e0>, 946021.36840352)]']connector: <aiohttp.connector.TCPConnector object at 0x7ffa30df84d0>

I tried adding below code but still getting error:
```python
await client._session.close()
```

#### Output of ``pip freeze`` 

<details>

azure-functions==1.11.2
pyodbc==4.0.32
pandas==1.3.4
azure-identity==1.6.1
azure-kusto-data[aio]==3.1.0
azure-storage-file-datalake==12.5.0
packaging==21.2
stringcase==1.2.0
aiohttp==3.8.1

</details>

AsafMah:
Can't recreate - can you give a more full code sample, including what reactor you use to run the async method?

AsafMah:
Closing as there is no response in a few months.

jamestrousdale:
@AsafMah Would you please re-open this?

I think the missing context here is that it does not happen with all authentication methods.

I am currently experiencing exactly this issue with managed identity (workload identity) authentication. Could you see if you could reproduce with that authentication method?

AsafMah:
Thank you for the in depth research!

I'll work on a fix

AsafMah:
Fixed in 4.3.0 - let us know if the issue persists

PR #513, opened 2023-11-30
Fix closing of async token providers

AsafMah:
### Fixed
* Async token providers are properly closed (#417)
	*  This PR makes using sync clients with async token providers an error (since they can't be properly closed). Is this considered breaking or bug fix?

### DIFF
--- a/azure-kusto-data/azure/kusto/data/_token_providers.py
+++ b/azure-kusto-data/azure/kusto/data/_token_providers.py
@@ -29,13 +29,10 @@ DeviceCallbackType = Callable[[str, str, datetime], None]
         If this argument isn't provided, the credential will print instructions to stdout."""
 
 try:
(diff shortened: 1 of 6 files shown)


azure-kusto-data/azure/kusto/data/_token_providers.py
```diff
@@ -201,6 +204,12 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    async def __aenter__(self):
```
ohadbitt, line 207: This is only for tests?

ohadbitt (approved):

azure-kusto-data/azure/kusto/data/_token_providers.py
AsafMah: I guess - but it's weird to implement aexit without aenter, they usually go together

PR #517, opened 2024-01-08
Streaming from blob

ohadbitt:
### Added
- Streaming ingestion from blob
### Fixed
- Managed streaming fallback to queued


azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -199,7 +199,7 @@ def execute_streaming_ingest(
         self,
```
AsafMah, line 199: As always, update CHANGELOG.md

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -199,7 +199,7 @@ def execute_streaming_ingest(
...
         database: Optional[str],
         table: str,
-        stream: IO[AnyStr],
+        data: Union[IO[AnyStr], str],
```
AsafMah, line 202: Not a big fan of overloading this argument.
Why not add another parameter for blob uri? Or a different method?

Even if not so, from looking at the type it seems like it would ingest what's in the str, not a blob that the str points to, you might want a better class.

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -295,7 +297,7 @@ def _execute(
             endpoint,
             headers=request_headers,
             json=json_payload,
```
AsafMah, line 299: Wait, up until now we sent BOTH json and data? that doesn't seem right

azure-kusto-data/azure/kusto/data/client_base.py
```diff
@@ -128,9 +128,9 @@ class ExecuteRequestParams:
     def __init__(
```
AsafMah, line 128: This is getting pretty messy.
Maybe split it to multiple ctors, each with the params they need:
from_query, from_mgmt, from_streaming_ingest_reader, from_streaming_ingest_uri or something

azure-kusto-ingest/azure/kusto/ingest/descriptors.py
```diff
@@ -150,6 +152,10 @@ def get_tracing_attributes(self) -> dict:
             obfuscated_path = self.path.split("?")[0].split(";")[0]
         return {self._BLOB_URI: obfuscated_path, self._SOURCE_ID: str(self.source_id)}
 
+    def fillSize(self):
```
AsafMah, line 155: pythonic - fill_size

azure-kusto-ingest/azure/kusto/ingest/managed_streaming_ingest_client.py
```diff
@@ -108,30 +108,20 @@ def ingest_from_stream(self, stream_descriptor: Union[StreamDescriptor, IO[AnySt
         stream = stream_descriptor.stream
 
         buffered_stream = read_until_size_or_end(stream, self.MAX_STREAMING_SIZE_IN_BYTES + 1)
-
-        if len(buffered_stream.getbuffer()) > self.MAX_STREAMING_SIZE_IN_BYTES:
-            stream_descriptor.stream = chain_streams([buffered_stream, stream])
-            return self.queued_client.ingest_from_stream(stream_descriptor, ingestion_properties)
+        length = len(buffered_stream.getbuffer())
```
AsafMah, line 111: Why was this removed?

azure-kusto-ingest/azure/kusto/ingest/managed_streaming_ingest_client.py
```diff
@@ -151,9 +141,37 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
...
+        for attempt in Retrying(stop=stop_after_attempt(self._num_of_attempts), wait=wait_random_exponential(max=self._max_seconds_per_retry), reraise=True):
+            with attempt:
+                client_request_id = ManagedStreamingIngestClient._get_request_id(descriptor.source_id, attempt.retry_state.attempt_number - 1)
+                # trace attempt to ingest from stream
+                if isinstance(descriptor, StreamDescriptor):
+                    descriptor.stream.seek(0, SEEK_SET)
+                    invoker = lambda: self.streaming_client._ingest_from_stream_with_client_request_id(descriptor, props, client_request_id)
+                else:
+                    invoker = lambda: self.streaming_client.ingest_from_blob(descriptor, props, client_request_id)
+                return MonitoredActivity.invoke(invoker, name_of_span="ManagedStreamingIngestClient.ingest_from_stream_attempt")
```
AsafMah, line 173: Add the attempt number, and other things as a param?

AsafMah (changes requested):

azure-kusto-data/azure/kusto/data/client.py
ohadbitt: no, up until now you had only one of them at the time, now I have both

azure-kusto-data/azure/kusto/data/client.py
```diff
@@ -191,15 +200,25 @@ def execute_mgmt(self, database: Optional[str], query: str, properties: Optional
...
+        return self._execute(self._mgmt_endpoint, request, properties)
 
     @distributed_trace(name_of_span="KustoClient.streaming_ingest", kind=SpanKind.CLIENT)
     def execute_streaming_ingest(
         self,
         database: Optional[str],
         table: str,
-        stream: IO[AnyStr],
+        stream: Optional[IO[AnyStr]],
+        blob_url: Optional[str],
```
AsafMah, line 221: Technically a breaking change - since this function is "public" +
We probably need to reflect the change in the async version, even if we don't use it in ingest (since they need to have api compatibility )

azure-kusto-ingest/tests/test_e2e_ingest.py
```diff
@@ -538,3 +538,30 @@ async def test_streaming_ingest_from_dataframe(self):
         self.ingest_client.ingest_from_dataframe(df, ingestion_properties)
 
         await self.assert_rows_added(1, timeout=120)
+
+    @pytest.mark.asyncio
+    def test_streaming_ingest_from_blob(self, is_managed_streaming):
```
AsafMah, line 543: Test for fallback to queued (specifically if the size is too big)

AsafMah (changes requested):

azure-kusto-data/azure/kusto/data/aio/client.py
```diff
@@ -118,41 +164,34 @@ async def execute_streaming_query(
...
-        )
-        json_payload = request_params.json_payload
-        request_headers = request_params.request_headers
-        timeout = request_params.timeout
+
+        request_headers = request.request_headers
+        timeout = request.timeout
         if self._aad_helper:
             request_headers["Authorization"] = await self._aad_helper.acquire_authorization_header_async()
+        print("request headers: " + str(request_headers))
```
AsafMah, line 180: Remove this

AsafMah (approved):
One fix

PR #519, opened 2024-01-15
use new unittest.mock everywhere

a-detiste:
It is already used in one location:

`azure-kusto-data/tests/aio/test_kusto_client.py:from unittest.mock import patch`

The old 'mock' module is deprecated


a-detiste:
@microsoft-github-policy-service agree

AsafMah:
Thank you! approved

PR #526, opened 2024-02-29
App and user for tracing

t-ronmoneta:
### Added
- app and user for tracing fields in ingestion properties
- Took those fields from KCSB and added them to ingestion Propeties in queued ingest client

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
@@ -49,7 +49,7 @@ class QueuedIngestClient(BaseIngestClient):
         self._endpoint_service_type = None
         self._suggested_endpoint_uri = None
         self.application_for_tracing = kcsb.application_for_tracing
-        self.user_for_tracing = kcsb.user_name_for_tracing
+        self.client_version_for_tracing = kcsb.client_details
 
     def close(self) -> None:
         self._resource_manager.close()
@@ -69,7 +69,7 @@ class QueuedIngestClient(BaseIngestClient):
         """
         file_descriptor = FileDescriptor.get_instance(file_descriptor)
         ingestion_properties.application_for_tracing = self.application_for_tracing
-        ingestion_properties.user_for_tracing = self.user_for_tracing
+        ingestion_properties.client_version_for_tracing = self.client_version_for_tracing
         IngestTracingAttributes.set_ingest_descriptor_attributes(file_descriptor, ingestion_properties)
 
         super().ingest_from_file(file_descriptor, ingestion_properties)
@@ -98,7 +98,7 @@ class QueuedIngestClient(BaseIngestClient):
         """
         stream_descriptor = StreamDescriptor.get_instance(stream_descriptor)
         ingestion_properties.application_for_tracing = self.application_for_tracing
-        ingestion_properties.user_for_tracing = self.user_for_tracing
+        ingestion_properties.client_version_for_tracing = self.client_version_for_tracing
         IngestTracingAttributes.set_ingest_descriptor_attributes(stream_descriptor, ingestion_properties)
 
         super().ingest_from_stream(stream_descriptor, ingestion_properties)
--- a/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
@@ -48,8 +48,8 @@ class QueuedIngestClient(BaseIngestClient):
         self._resource_manager = _ResourceManager(KustoClient(kcsb))
         self._endpoint_service_type = None
         self._suggested_endpoint_uri = None
-        self.application_for_tracing = kcsb.application_for_tracing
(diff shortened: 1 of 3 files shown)


azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -66,6 +68,8 @@ def ingest_from_file(self, file_descriptor: Union[FileDescriptor, str], ingestio
         :param azure.kusto.ingest.IngestionProperties ingestion_properties: Ingestion properties.
         """
         file_descriptor = FileDescriptor.get_instance(file_descriptor)
+        ingestion_properties.application_for_tracing = self.application_for_tracing
```
AsafMah, line 71: Code duplication between the two methods

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -66,6 +68,8 @@ def ingest_from_file(self, file_descriptor: Union[FileDescriptor, str], ingestio
...
+        ingestion_properties.user_for_tracing = self.user_for_tracing
```
AsafMah, line 72: What if the user provided their own values in the ingestion props? it will write over it

azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
```diff
@@ -218,6 +220,8 @@ def __init__(
         self.report_method = report_method
         self.validation_policy = validation_policy
         self.additional_properties = additional_properties
+        self.application_for_tracing = application_for_tracing
```
AsafMah, line 223: Unless I'm missing something, you don't forward it to IngestionBlobInfo, so this goes nowhere

AsafMah (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
t-ronmoneta: @yogilad unless i didnt understand correctly, it should be only given from KCSB, and not from user, right? (in which case i do need to remove it as input variables in _init_ im aware)

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
t-ronmoneta: shouldnt i set this no matter which option the user chose? and i cant just put it at the end in ingest_from_blob once because the ingestion properties are handled beforehand (and i dont know in what phase the DM will need it)

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: You can overwrite whatever the user wrote there

azure-kusto-ingest/azure/kusto/ingest/ingestion_properties.py
yogilad: Why not copy ctoring it ?

t-ronmoneta: see,s like ingestionblobinfo handles ingestionproperties field differently for each field, so not sure we just wanna copy it

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
t-ronmoneta: @yogilad @AsafMah Asaf disagrees, should we discuss it for a moment?

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -48,6 +48,8 @@ def __init__(self, kcsb: Union[str, KustoConnectionStringBuilder]):
         self._resource_manager = _ResourceManager(KustoClient(kcsb))
         self._endpoint_service_type = None
         self._suggested_endpoint_uri = None
+        self.application_for_tracing = kcsb.application_for_tracing
+        self.user_for_tracing = kcsb.user_name_for_tracing
```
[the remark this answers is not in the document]
t-ronmoneta: We said we wanted the actual kcsb.client_details right? it is returned as ClientDetails(self.application_for_tracing, self.user_name_for_tracing) - which is not JSON serializable. do we perhaps want to extract the inner data first into a tuple, and use that?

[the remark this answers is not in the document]
yogilad: Is it a class with accessible properties?
If so, simply copy the right ones to properties of the queued message

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
yogilad: We discussed this yesterday.
The conclusion was this needs to be overwritten and possibly on the message being queued to the service (blob info or something) and not the ingestion properties themselves.

this also has the benefit of not changing the object the user passed to the API call

AsafMah (dismissed):
Great!

PR #527, opened 2024-03-21
Small fixes

AsafMah:
- Fixed not correctly awaiting for async token providers to close
- Tests fixes:
  - Use new method for the event loop fixed, since the old one is deprecated
  - Fixed tests not being awaited correctly
  - Fixed exceptions if only some of the env vars are provided

### DIFF
--- a/azure-kusto-data/tests/test_e2e_data.py
+++ b/azure-kusto-data/tests/test_e2e_data.py
@@ -22,6 +22,11 @@ from azure.kusto.data.exceptions import KustoServiceError
 from azure.kusto.data.kusto_trusted_endpoints import MatchRule, well_known_kusto_endpoints
 from azure.kusto.data.streaming_response import FrameType
 
+@pytest.fixture(scope="module")
+def event_loop_policy(request):
+    if platform.system() == "Windows":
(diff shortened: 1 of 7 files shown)


azure-kusto-data/tests/test_e2e_data.py
```diff
@@ -23,6 +23,13 @@
 from azure.kusto.data.streaming_response import FrameType
 
 
+@pytest.fixture(scope="module")
+def event_loop_policy(request):
+    if platform.system() == "Windows":
+        return asyncio.WindowsSelectorEventLoopPolicy()
```
yihezkel, line 29: I'm curious why use `WindowsSelectorEventLoopPolicy` for Windows, instead of always using `DefaultEventLoopPolicy`?

(And if we always used `DefaultEventLoopPolicy`, there'd be no need for this method altogether.)

yihezkel (approved):

azure-kusto-data/tests/test_e2e_data.py
AsafMah: It's a problem upstream with pytest-asyncio related to closing the loop when testing ends. Then I think they fixed it and then there was a regression, so I added a comment and leaving it for now.

PR #529, opened 2024-04-03
pandas

AsafMah:

### DIFF
--- a/azure-kusto-data/azure/kusto/data/helpers.py
+++ b/azure-kusto-data/azure/kusto/data/helpers.py
@@ -48,9 +48,13 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
 
     columns = [col.column_name for col in table.columns]
     frame = pd.DataFrame(table.raw_rows, columns=columns)
+    print (frame.dtypes)
+    print (frame.infer_objects().dtypes)
 
     # fix types
     for col in table.columns:
+        if col.column_type == "string" and hasattr(pd, "StringDType"):
+            frame[col.column_name] = frame[col.column_name].astype(pd.StringDType())
         if col.column_type == "bool":
             frame[col.column_name] = frame[col.column_name].astype(pd.BooleanDtype() if nullable_bools else bool)
         elif col.column_type == "int":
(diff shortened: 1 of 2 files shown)


azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -48,9 +48,13 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
 
     columns = [col.column_name for col in table.columns]
     frame = pd.DataFrame(table.raw_rows, columns=columns)
+    print(frame.dtypes)
+    print(frame.infer_objects().dtypes)
```
yihezkel, line 52: Are these prints in the non-test code intentional?

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -48,9 +48,13 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
 
     # fix types
     for col in table.columns:
+        if col.column_type == "string" and hasattr(pd, "StringDType"):
```
yihezkel, line 56: Suggest adding a check on the return value as well, just in case.

yihezkel (approved):
Approved with suggestions

azure-kusto-data/azure/kusto/data/helpers.py
AsafMah: Thanks!

azure-kusto-data/azure/kusto/data/helpers.py
AsafMah: On what? the return value of the type? not sure how?

yihezkel: I think once `hasattr(pd, "StringDType")` is `True`, you can check its value via `pd.StringDType`. It isn't critical, but this can help ensure the type is not only defined on the `df`, but as expected, which could theoretically happen as the external package evolves.

(I'd expect the value to be `string`, but I could be wrong.)

AsafMah: It's not a string, the type *is* pd.StringDType, so I can really only compare it to itself..

PR #530, opened 2024-04-08
App and user for tracing

t-ronmoneta:
### Changed
- removed tracing params from ingestion properties

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
@@ -49,7 +49,7 @@ class QueuedIngestClient(BaseIngestClient):
         self._endpoint_service_type = None
         self._suggested_endpoint_uri = None
         self.application_for_tracing = kcsb.application_for_tracing
-        self.user_for_tracing = kcsb.user_name_for_tracing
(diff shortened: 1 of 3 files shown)


azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
```diff
@@ -135,7 +135,13 @@ def ingest_from_blob(self, blob_descriptor: BlobDescriptor, ingestion_properties
         queues = self._resource_manager.get_ingestion_queues()
 
         authorization_context = self._resource_manager.get_authorization_context()
-        ingestion_blob_info = IngestionBlobInfo(blob_descriptor, ingestion_properties=ingestion_properties, auth_context=authorization_context)
+        ingestion_blob_info = IngestionBlobInfo(
```
AsafMah, line 138: I don't think you did a complete job - see lines 128 and 129

AsafMah (changes requested):

azure-kusto-ingest/azure/kusto/ingest/ingest_client.py
t-ronmoneta: i deleted them and dont have them in the local version, im not really sure what happend here

AsafMah (approved):

PR #531, opened 2024-04-21
Fix headers escaping and parsing

AsafMah:
Changes the tracing format to be always allowed in headers, and easier to reason about:

- Only allowed chars: [a-zA-Z0-9_.\-()] (Note: not using \w because we don't want non-ascii chars)
 - Escape every part of the header, even when using set_connector_details 
- Remove the brackets around escaped fields, as they are not needed to parse anymore. So - Parsing is now simply splitting by |, and then you have pairs of key:value. 
- Limit the length of each field to 100 

So old: Kusto.Café:{1 . 0}|App.{Something}:{weird version} 
New: Kusto.Caf_:1_._0|App.Something:weird_versio

### DIFF
--- a/azure-kusto-data/azure/kusto/data/client_details.py
+++ b/azure-kusto-data/azure/kusto/data/client_details.py
@@ -13,6 +13,7 @@ NONE = "NONE"
 REPLACE_REGEX = re.compile(r"[^a-zA-Z0-9_.\-()]")
 MAX_HEADER_LENGTH = 100
 
+
 @functools.lru_cache(maxsize=1)
 def default_script() -> str:
     """Returns the name of the script that is currently running"""
--- a/azure-kusto-data/azure/kusto/data/client_details.py
+++ b/azure-kusto-data/azure/kusto/data/client_details.py
@@ -8,10 +8,10 @@ from typing import List, Tuple, Optional
 from .env_utils import get_env
 from azure.kusto.data._version import VERSION
 
-NONE = "[none]"
+NONE = "NONE"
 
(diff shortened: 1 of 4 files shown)


azure-kusto-data/azure/kusto/data/client_details.py
```diff
@@ -80,17 +85,17 @@ def set_connector_details(
...
+        params.append(("App." + app_name, app_version))
         params.extend(additional_fields or [])
 
         user = NONE
 
         if send_user:
             user = override_user or default_user()
 
-        return ClientDetails(application_for_tracing=format_header(params), user_name_for_tracing=user)
+        return ClientDetails(application_for_tracing=format_header(params), user_name_for_tracing=user, should_escape=False)
```
ohadbitt, line 101: should_escape = true ?

azure-kusto-data/azure/kusto/data/client_details.py
```diff
@@ -57,18 +58,22 @@ def format_header(args: List[Tuple[str, str]]) -> str:
...
 
     def __post_init__(self):
         self.application_for_tracing = self.application_for_tracing or default_script()
         self.user_name_for_tracing = self.user_name_for_tracing or default_user()
+        if self.should_escape:
+            self.application_for_tracing = escape_field(self.application_for_tracing)
+            self.user_name_for_tracing = escape_field(self.user_name_for_tracing)
 
     @staticmethod
     def set_connector_details(
```
ohadbitt, line 79: why does a static method creating details is called set_connector_details

PR #533, opened 2024-05-05
users/asafsmahlev/minor_fixes

### LINKED ISSUE #532: KustoAsyncUsageError is thrown every time when using managed identity authentication method
#### Code Sample, a copy-pastable example if possible

```python
with KustoClient(query_connection_string_builder) as kusto_client:
        with QueuedIngestClient(ingest_connection_string_builder) as queued_ingest_client:
               .....

```
#### Problem description
The above code results in
```
KustoAsyncUsageError: Method Can't close async token provider with sync close can't be called from a synchronous client
```
when using `KustoConnectionStringBuilder.with_aad_managed_service_identity_authentication()` method for creating the connection string.

The bug was introduced in version 4.3.1 with the following code in `azure-kusto-data/azure/kusto/data/_token_providers.py`:
image

Commit: https://github.com/Azure/azure-kusto-python/commit/cd0ee8a3856e585e98199dbd94a477bbc2438197

#### If query related, does it happen on other platforms (Kusto Web UI, Kusto Explorer)? 

[this step is to help pin point problems that are only specific to this platform.]

#### Output of ``pip freeze`` 

<details>

[paste the output of ``pip freeze`` here below this line]
azure-kusto-data==4.4.0
azure-kusto-ingest==4.4.0

</details>

AsafMah:
Fixes #532

### DIFF
--- a/azure-kusto-data/tests/test_token_providers.py
+++ b/azure-kusto-data/tests/test_token_providers.py
@@ -370,3 +370,21 @@ class TokenProviderTests(unittest.TestCase):
         ) as provider:
             token = provider.get_token()
             assert TokenProviderTests.get_token_value(token) is not None
+
+    @staticmethod
+    def test_sanity_close():
+        DUMMY_URI = "https://dummy_uri.kusto.windows.net"
+        providers = [
+            MsiTokenProvider(DUMMY_URI),
+            AzCliTokenProvider(DUMMY_URI),
+            UserPassTokenProvider(DUMMY_URI, "organizations", "a", "b"),
+            ApplicationKeyTokenProvider(DUMMY_URI, "a", "b", "c"),
+            ApplicationCertificateTokenProvider(DUMMY_URI, "a", "b", "c", "d"),
+            CallbackTokenProvider(lambda: "a", None),
+            AzureIdentityTokenCredentialProvider(DUMMY_URI, credential=DefaultAzureCredential()),
+            DeviceLoginTokenProvider(DUMMY_URI, "organizations", lambda x, x2, x3: None),
+        ]
+
+        for provider in providers:
+            with provider:
+                pass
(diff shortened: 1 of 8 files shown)


azure-kusto-data/tests/test_token_providers.py
```diff
@@ -370,3 +370,21 @@ def test_azure_identity_default_token_provider():
...
+            ApplicationKeyTokenProvider(DUMMY_URI, "a", "b", "c"),
+            ApplicationCertificateTokenProvider(DUMMY_URI, "a", "b", "c", "d"),
+            CallbackTokenProvider(lambda: "a", None),
+            AzureIdentityTokenCredentialProvider(DUMMY_URI, credential=DefaultAzureCredential()),
+            DeviceLoginTokenProvider(DUMMY_URI, "organizations", lambda x, x2, x3: None),
+        ]
+
+        for provider in providers:
+            with provider:
+                pass
```
yihezkel, line 390: Not in scope for this PR, but there's so much duplicated code between the aio and non-async code.
Isn't there a way we could dedupe most of it?

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
```diff
@@ -65,9 +65,11 @@ def __init__(self, status: IngestionStatus, database: str, table: str, source_id
 
     def __repr__(self):
         # Remove query parameters from blob_uri, if exists
+        obfuscated_path = None
         if isinstance(self.blob_uri, str):
             obfuscated_path = self.blob_uri.split("?")[0].split(";")[0]
-        return f"IngestionResult(status={self.status}, database={self.database}, table={self.table}, source_id={self.source_id}, blob_uri={obfuscated_path})"
+        blob_uri = f", blob_uri={obfuscated_path}" if obfuscated_path else ""
```
yihezkel, line 71: Do we want to indicate in the text that the URL is obfuscated?

yihezkel (approved):

azure-kusto-data/tests/test_token_providers.py
AsafMah: We were looking into it. There is some things we can do, but we are waiting for some "big" solution

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
AsafMah: Good idea

PR #535, opened 2024-05-08
raise NotImplementedError if ignore_first_record is True

JamieHolloway:
### Added
Added NotImplementedError if ignore_first_record is True to avoid introducing data integrity error when loading .csv file when invoking method ingest_from_file.  if ignore_first_record is set to True, then the header/first-row of the .csv file is ingested.  This change prevents the introduction of a data integrity error until this feature is implemented.  I tested change, but did not execute full regression.


AsafMah:
For now, this change will be breaking for people who share their ingestion properties between clients.

We are in the process of refactoring ingestion on a bigger scale, that will also solve issues like this.

Additionally, at some point this will be added to the service.

So at this point, I'm closing the PR.

Issue #512, opened 2023-11-29
Enhance Dataframe Conversion to Support All Float Data Types (float16, float32, float64) in `dataframe_from_result_table()`


thlautenschlaeger:
**Is your feature request related to a problem? Please describe.**
Currently, the dataframe_from_result_table function in the azure-kusto-python library converts float data types to float64 by default. However, there are scenarios where using float16 or float32 might be more appropriate, such as when working with large datasets where memory efficiency is crucial, or when interfacing with systems that primarily use float32 data types.

**Describe the solution you'd like**
I propose extending the [dataframe_from_result_table()](https://github.com/Azure/azure-kusto-python/blob/master/azure-kusto-data/azure/kusto/data/helpers.py#L31-L67) function to allow the user to specify the float data type (float16, float32, float64). This can be achieved by adding an optional parameter, such as float_dtype, with a default value of float64. The function would then use this parameter to determine the data type for floating-point columns in the resulting DataFrame.

**Describe alternatives you've considered**
An alternative would be to create a separate function specifically for float32 conversion, but this approach seems redundant and less efficient. Allowing the user to specify the float data type in the existing function provides a more elegant and flexible solution.

**Additional context**
This enhancement would increase the usability of the azure-kusto-python library for a broader range of applications, particularly those dealing with large datasets or specific system requirements. It also aligns with the principles of flexibility and user choice in software design. I have already implemented a version of this feature in my local environment and found it to be extremely useful for my use cases.

thlautenschlaeger:
I'd like to contribute this feature.

AsafMah:
Please do, let us know if you have any questions

AsafMah:
Are you still interested in implementing this?

AsafMah:
Added in v4.5.0.

Thanks @mayamarom10 !

PR #538, opened 2024-05-30
Changed dataframe_from_result_table function 

mayamarom10:
**Changed**: #512
Changed the dataframe_from_result_table function - added dictionary options as parameters for matching python types to pandas types by column name or by python type.


azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -87,3 +89,23 @@ def get_string_tail_lower_case(val, length):
         return val.lower()
 
     return val[len(val) - length :].lower()
+
+
+def real_or_decimal(frame, col):
```
AsafMah, line 94: Not yours, but: this can actually change the value of decimals, this is the wrong type to parse to (though changing it would be breaking, we need to think about that)

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -87,3 +89,23 @@ def get_string_tail_lower_case(val, length):
...
+    frame[col] = frame[col].replace("NaN", np.NaN).replace("Infinity", np.PINF).replace("-Infinity", np.NINF)
+    frame[col] = pd.to_numeric(frame[col], errors="coerce").astype(pd.Float64Dtype())
+    return frame[col]
+
+
+def datetime_func(frame, col):
```
AsafMah, line 100: parse_datetime ?

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,5 +1,7 @@
 import sys
 from typing import TYPE_CHECKING, Union
+import numpy as np
```
AsafMah, line 3: Remove these - before this we were careful to not just import these modules, since (for now) the pandas feature is optional, and users might not have these packages.

azure-kusto-data/tests/input/dataframe.json
```diff
@@ -49,6 +49,19 @@
         "ColumnName": "RecordInt",
         "ColumnType": "int"
     },
+        {
```
AsafMah, line 52: formatting

azure-kusto-data/tests/input/dataframe.json
```diff
@@ -49,6 +49,19 @@
...
+            "ColumnName": "RecordInt64",
+            "ColumnType": "int"
+        },
+        {
+            "ColumnName": "RecordLong",
+            "ColumnType": "long"
+        },
+
+        {
+            "ColumnName": "RecordGUID",
```
AsafMah, line 62: Good that you added more types, but while we're at it - add all of them?

azure-kusto-data/azure-kusto-data.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
AsafMah, line 1: These files are unrelated, and usually you won't commit them, but since they will help intellij automatically sort out the module issue for people it might worth it to actually commit them

azure-kusto-data/tests/test_helpers.py
```diff
@@ -18,19 +18,32 @@ def test_dataframe_from_result_table():
         data = response_file.read()
 
     response = KustoResponseDataSetV2(json.loads(data))
-    df = dataframe_from_result_table(response.primary_results[0])
+    test_dict1 = {"RecordName": lambda col, frame: frame[col].astype("str"), "RecordInt64": lambda col, frame: frame[col].astype("int64")}
+    test_dict2 = {"int": lambda col, frame: frame[col].astype("int32")}
```
AsafMah, line 22: Not sure about this test - doesn't int already get converted to int32? try a completely different type

AsafMah (changes requested):
First review, don't worry, they will get shorter.

azure-kusto-data/azure/kusto/data/helpers.py
mayamarom10: I don't know about a different pandas type that would be a better match

azure-kusto-data/tests/test_helpers.py
mayamarom10: I wanted to test that the conversion of int types to int32 won't override the conversion by column name to int64. I'll add another test that converts to a completely different type.

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -29,13 +31,16 @@ def to_pandas_timedelta(raw_value: Union[int, float, str]) -> "pandas.Timedelta"
             return pd.to_timedelta(formatted_value)
 
 
-def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingResultTable]", nullable_bools: bool = False) -> "pandas.DataFrame":
+def dataframe_from_result_table(
+    table: "Union[KustoResultTable, KustoStreamingResultTable]", nullable_bools: bool = False, column_name_totype_dict: dict = {}, type_totype_dict: dict = {}
```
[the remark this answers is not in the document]
AsafMah: Technically the converters can do a lot more than just change the type since it's a lambda - they can filter them, format them, etc. So I think the general name is good (+ shorter)

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -49,32 +54,29 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
+        "long": lambda col, df: df[col].astype(pd.Int64Dtype()),
+        "real": lambda col, df: real_or_decimal(df, col),
+        "decimal": lambda col, df: real_or_decimal(df, col),
+        "datetime": lambda col, df: datetime_func(df, col),
+        "timespan": lambda col, df: df[col].apply(to_pandas_timedelta),
+    }
+
+    by_name_bool = True if column_name_totype_dict else False
+    by_type_bool = True if type_totype_dict else False
+
```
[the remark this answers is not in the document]
AsafMah: If you do this you only need one dict access, but maybe it's a bit of a waste.

`default_dict.get(type) or user_dict.get(type)` would work too.

azure-kusto-data/azure-kusto-data.iml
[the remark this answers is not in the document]
AsafMah: See my other comment - it might be beneficial to keep it, to make the setup easier.

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,10 +1,30 @@
+from __future__ import annotations
```
AsafMah, line 1: Don't use this - it breaks libraries. Just you quote your annotations ("pd.Timedelta")

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,10 +1,30 @@
...
-from typing import TYPE_CHECKING, Union
+import types
+from typing import TYPE_CHECKING, Union, Callable
 
 if TYPE_CHECKING:
-    import pandas
+    import pandas as pd
     from azure.kusto.data._models import KustoResultTable, KustoStreamingResultTable
 
+try:
```
AsafMah, line 10: I think we can avoid this.

You can

```python
from functools import lru_cache

@lru_cache
def default_dict() -> ...:
    return {...}

```


This will cache the result and you don't have to do the import hack.

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,10 +1,30 @@
+from __future__ import annotations
 import sys
-from typing import TYPE_CHECKING, Union
+import types
+from typing import TYPE_CHECKING, Union, Callable
 
 if TYPE_CHECKING:
-    import pandas
+    import pandas as pd
```
AsafMah, line 7: This change means you should change all the "pandas.X" types to "pd.X"

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,10 +1,30 @@
+from __future__ import annotations
 import sys
```
AsafMah, line 2: not used anymore

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -29,13 +49,22 @@ def to_pandas_timedelta(raw_value: Union[int, float, str]) -> "pandas.Timedelta"
             return pd.to_timedelta(formatted_value)
 
 
-def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingResultTable]", nullable_bools: bool = False) -> "pandas.DataFrame":
-    """Converts Kusto tables into pandas DataFrame.
+def dataframe_from_result_table(
+    table: "Union[KustoResultTable, KustoStreamingResultTable]",
+    nullable_bools: bool = False,
+    converters_by_type: dict[str, Callable[[str, "pandas.DataFrame"], "pandas.Series"]] = None,
```
AsafMah, line 55: the dict[...] syntax is python 3.10+
You need to import Dict from typings and do Dict[...]

AsafMah, line 55: Since they can be None, the type is actually Optional[Dict[...]]

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -29,13 +49,22 @@ def to_pandas_timedelta(raw_value: Union[int, float, str]) -> "pandas.Timedelta"
...
+    converters_by_column_name: dict[str, Callable[[str, "pandas.DataFrame"], "pandas.Series"]] = None,
```
AsafMah, line 56: This type is quite long, define an alias (at the top level)

Converter = Dict[str, Callable[[str, "pd.DataFrame"], "pd.Series"]]

azure-kusto-data/azure/kusto/data/helpers.py
AsafMah: If you keep the import anyway, put it before the TYPE_CHECKING if

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -49,32 +78,25 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
     columns = [col.column_name for col in table.columns]
     frame = pd.DataFrame(table.raw_rows, columns=columns)
 
-    # fix types
+    # converters_by_type overrides the default
+    if converters_by_type and not nullable_bools:
```
AsafMah, line 82: Ohad mentioned that creating a new dict is wasteful, and I have to agree.

So checking them one after the other is probably the correct way.
Sorry fot the back and forth.

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -49,32 +78,25 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
-        elif col.column_type == "real" or col.column_type == "decimal":
-            frame[col.column_name] = frame[col.column_name].replace("NaN", np.NaN).replace("Infinity", np.PINF).replace("-Infinity", np.NINF)
-            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors="coerce").astype(pd.Float64Dtype())
-        elif col.column_type == "datetime":
-            # Pandas before version 2 doesn't support the "format" arg
-            args = {}
-            if pd.__version__.startswith("2."):
-                args = {"format": "ISO8601", "utc": True}
+        if converters_by_column_name and col.column_name in converters_by_column_name.keys():
+            if isinstance(converters_by_column_name[col.column_name], types.LambdaType):
```
AsafMah, line 93: This is flimsy, if the users provide a function for example it won't do as expected. On the other hand, I don't think there is a common base for python types and pandas types, 

So with that I think we should remove the "giving just a type" functionality and only keep the converter lambdas

AsafMah: @ohadbitton maybe you have an idea

azure-kusto-data/azure/kusto/data/helpers.py
AsafMah: I guess changing it now would be breaking. I think when we are at exclusively at pandas 2 we can solve this.

AsafMah (changes requested):
More commentsj

azure-kusto-data/azure/kusto/data/helpers.py
mayamarom10: any concern in just checking if its a string type instead?

AsafMah: Seems like a good compromise. 

Make sure of course to change the type of the converters to `Dict[str, Union[str, Callable[[str, "pd.DataFrame"], "pd.Series"]]]`

azure-kusto-data/tests/test_helpers.py
```diff
@@ -11,28 +12,51 @@
 import numpy
 
 
-def test_dataframe_from_result_table():
-    """Test conversion of KustoResultTable to pandas.DataFrame, including fixes for certain column types"""
+class PandasTests(unittest.TestCase):
```
ohadbitt, line 15: Rename to HelpersTests (this will also align with the file name)

azure-kusto-data/tests/test_helpers.py
```diff
@@ -11,28 +12,51 @@
...
+    def test_dataframe_from_result_table(self):
+        """Test conversion of KustoResultTable to pandas.DataFrame, including fixes for certain column types"""
 
     with open(os.path.join(os.path.dirname(__file__), "input", "dataframe.json"), "r") as response_file:
         data = response_file.read()
 
     response = KustoResponseDataSetV2(json.loads(data))
-    df = dataframe_from_result_table(response.primary_results[0])
+    # Test that receives both types of dictionary parameters. Test that type conversion doesn't override column name conversion
```
ohadbitt, line 23:
```suggestion
    # Test when given both types of dictionary parameters that type conversion doesn't override column name conversion
```

azure-kusto-data/tests/test_helpers.py
```diff
@@ -11,28 +12,51 @@
...
+    test_dict_by_name = {"RecordName": lambda col, frame: frame[col].astype("str"), "RecordInt64": lambda col, frame: frame[col].astype("int64")}
+    test_dict_by_type = {"int": lambda col, frame: frame[col].astype("int32")}
+    test_int_to_float = {"int": "float64"}
+    df = dataframe_from_result_table(response.primary_results[0], converters_by_type=test_dict_by_type, converters_by_column_name=test_dict_by_name)
+    df_int_to_float = dataframe_from_result_table(response.primary_results[0], converters_by_type=test_int_to_float)
```
ohadbitt, line 28: Move df_int_to_float initialization to right before where it used (also test_int_to_float)

ohadbitt (approved):

azure-kusto-data/tests/test_helpers.py
AsafMah, line 15: I just noticed you created a class for this test - you don't need to do this. Notice how the rest of the tests don't have classes.
We use pytest for tests instead of unittest, so don't import/use it

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,14 +1,37 @@
...
-    import pandas
+    import pandas as pd
     from azure.kusto.data._models import KustoResultTable, KustoStreamingResultTable
 
+# Alias for dataframe_from_result_table converter type
+Converter = Optional[Dict[str, Union[str, Callable[[str, "pd.DataFrame"], "pd.Series"]]]]
+
+
+@lru_cache(maxsize=1, typed=False)
+def default_dict() -> Dict[str, Callable[[str, "pd.DataFrame"], "pd.Series"]]:
```
AsafMah, line 15: Return `Converter`

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -1,14 +1,37 @@
...
+        "decimal": lambda col, df: parse_float(df, col),
+        "datetime": lambda col, df: parse_datetime(df, col),
+        "timespan": lambda col, df: df[col].apply(to_pd_timedelta),
+    }
+
 
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License
-def to_pandas_timedelta(raw_value: Union[int, float, str]) -> "pandas.Timedelta":
+def to_pd_timedelta(raw_value: Union[int, float, str]) -> "pd.Timedelta":
```
AsafMah, line 34: This should be called parse_timedelta and be with the rest of the similar methods at the bottom

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -29,13 +52,22 @@ def to_pandas_timedelta(raw_value: Union[int, float, str]) -> "pandas.Timedelta"
...
+def dataframe_from_result_table(
+    table: "Union[KustoResultTable, KustoStreamingResultTable]",
+    nullable_bools: bool = False,
+    converters_by_type: Converter = None,
+    converters_by_column_name: Converter = None,
+) -> "pd.DataFrame":
+    f"""Converts Kusto tables into pandas DataFrame.
     :param azure.kusto.data._models.KustoResultTable table: Table received from the response.
     :param nullable_bools: When True, converts bools that are 'null' from kusto or 'None' from python to pandas.NA. This will be the default in the future.
+    :param converters_by_type: If given, converts specified types to corresponding types, else uses {default_dict()}. The dictionary maps from kusto
```
AsafMah, line 64: Mention here that you can give type names as well as lambdas

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -48,33 +80,25 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
 
     columns = [col.column_name for col in table.columns]
     frame = pd.DataFrame(table.raw_rows, columns=columns)
+    default = default_dict()
+
+    if nullable_bools:
```
AsafMah, line 85: Don't copy the entire dict for that.

Just use an if when getting from the default dict

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -48,33 +80,25 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
-            frame[col.column_name] = frame[col.column_name].astype(pd.Int64Dtype())
-        elif col.column_type == "real" or col.column_type == "decimal":
-            frame[col.column_name] = frame[col.column_name].replace("NaN", np.NaN).replace("Infinity", np.PINF).replace("-Infinity", np.NINF)
-            frame[col.column_name] = pd.to_numeric(frame[col.column_name], errors="coerce").astype(pd.Float64Dtype())
-        elif col.column_type == "datetime":
-            # Pandas before version 2 doesn't support the "format" arg
-            args = {}
-            if pd.__version__.startswith("2."):
-                args = {"format": "ISO8601", "utc": True}
+        if converters_by_column_name and col.column_name in converters_by_column_name.keys():
```
AsafMah, line 90: You don't need .keys() for an "in" check

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -48,33 +80,25 @@ def dataframe_from_result_table(table: "Union[KustoResultTable, KustoStreamingRe
...
-        elif col.column_type == "timespan":
-            frame[col.column_name] = frame[col.column_name].apply(to_pandas_timedelta)
+                frame[col.column_name] = converters_by_column_name[col.column_name](col.column_name, frame)
+        elif converters_by_type and col.column_type in converters_by_type.keys():
+            if isinstance(converters_by_type[col.column_type], str):
+                frame[col.column_name] = frame[col.column_name].astype(converters_by_type[col.column_type])
+            else:
+                frame[col.column_name] = converters_by_type[col.column_type](col.column_name, frame)
+        else:
+            frame[col.column_name] = default[col.column_type](col.column_name, frame)
```
AsafMah, line 102: This code is a bit repetitive. You can have a much nicer loop if you separate the retrival with the usage:


```suggestion
    for col in table.columns:
        column_name = col.column_name
        column_type = col.column_type

        if converters_by_column_name and column_name in converters_by_column_name:
            converter = converters_by_column_name[column_name]
        elif converters_by_type and column_type in converters_by_type:
            converter = converters_by_type[column_type]
        elif nullable_bools and column_type == "bool":
                converter = lambda col, df: df[col].astype(pd.BooleanDtype())
        else:
            converter = default[column_type]

        if isinstance(converter, str):
            frame[column_name] = frame[column_name].astype(converter)
        else:
            frame[column_name] = converter(column_name, frame)
```

azure-kusto-data/azure-kusto-data.iml
```diff
@@ -0,0 +1,9 @@
...
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
```
AsafMah, line 5: Add the correct source roots

```suggestion
    <content url="file://$MODULE_DIR$">
      <sourceFolder url="file://$MODULE_DIR$/azure" isTestSource="false" />
      <sourceFolder url="file://$MODULE_DIR$/tests" isTestSource="true" />
    </content>
```

azure-kusto-ingest/azure-kusto-ingest.iml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
```
AsafMah, line 5:
```suggestion
    <content url="file://$MODULE_DIR$">
      <sourceFolder url="file://$MODULE_DIR$/azure" isTestSource="false" />
      <sourceFolder url="file://$MODULE_DIR$/tests" isTestSource="true" />
    </content>
    ```

File: quick_start/quick_start.iml [hunk 29]
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
```
AsafMah, line 5:
```suggestion
    <content url="file://$MODULE_DIR$">
      <sourceFolder url="file://$MODULE_DIR$/azure" isTestSource="false" />
      <sourceFolder url="file://$MODULE_DIR$/tests" isTestSource="true" />
    </content>
```

AsafMah (changes requested):

azure-kusto-data/azure/kusto/data/helpers.py
mayamarom10: Even though in this case the return type will never be None? (So its not Optional like Converter)

AsafMah: Great point!

I think you should define the Converter type without the Optional, and then annotate Optional[Converter] whenever it is optional (it's also clearer)

AsafMah (approved):

Issue #542, opened 2024-06-18
azure-kusto-data 4.5 helpers fail constructing dataframe of .show materialized-views


borlaugen:
```python
materialized_views_response = self.kusto_client.execute(
    self._database, ".show materialized-views"
)
materialized_tables = dataframe_from_result_table(
    materialized_views_response.primary_results[0]
)

```
The above code will with version 4.5 raise a  KeyError exception on key 'time' since the LookBack column has ColumnType 'time'.

This is most likely introduced by commit 4f9af0f5926d73412e8865d3edd5a7e469efee01

![image](https://github.com/Azure/azure-kusto-python/assets/59911423/2e69b203-32b3-47e1-8424-3af6cdb6faf3)

AsafMah:
Thanks, we will fix this.

As a workaround for now you can pass a converter like so:
```
            from azure.kusto.data.helpers import parse_timedelta
            data_frame = dataframe_from_result_table(data, converters_by_type={"time": lambda col, df: df[col].apply(parse_timedelta)})
```

AsafMah:
Fixed in 4.5.1

PR #543, opened 2024-06-18
Fix df from result table error message + old types 

mayamarom10:
fixed error message in case of none existing type 
added old type names in kusto to default dictionary #542


azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -67,16 +75,18 @@ def dataframe_from_result_table(
...
+            converter = converters_by_type.get(column_type)
         elif nullable_bools and column_type == "bool":
             converter = lambda col, df: df[col].astype(pd.BooleanDtype())
         else:
-            converter = default[column_type]
+            converter = default.get(column_type)
         if isinstance(converter, str):
             frame[column_name] = frame[column_name].astype(converter)
         else:
+            if converter is None:
```
AsafMah, line 88: move this before the isinstance

azure-kusto-data/azure/kusto/data/helpers.py
```diff
@@ -18,14 +18,22 @@ def default_dict() -> Converter:
     return {
```
AsafMah, line 18: Add a test with an alias name, and with a missing type name

AsafMah (changes requested):

azure-kusto-data/azure/kusto/data/helpers.py
ohadbitt, line 88: Move None check before isinstance

ohadbitt (dismissed):

azure-kusto-data/tests/test_helpers.py
```diff
@@ -90,10 +96,21 @@ def test_dataframe_from_result_table():
...
-df_int_to_float = dataframe_from_result_table(response.primary_results[0], converters_by_type=test_int_to_float)
+ignore_missing_type = {
+    "MissingType": lambda col, frame: frame[col].astype("str"),
+}
+df_int_to_float = dataframe_from_result_table(response.primary_results[0], converters_by_type=test_int_to_float, converters_by_column_name=ignore_missing_type)
 assert type(df_int_to_float.iloc[0].RecordInt) is numpy.float64
 assert df.iloc[0].RecordInt == 5678
 
+# Testing missing type conversion
+raised = False
```
AsafMah, line 107: Use pytest.raises

AsafMah (approved):

PR #545, opened 2024-07-02
Fix json 401 error

mayamarom10:
1. Updated handle_http_error func to handle 401 errors. 
2. Updates client.py to check if json is empty before parsing.


azure-kusto-ingest/tests/test_e2e_ingest.py
```diff
@@ -564,3 +565,11 @@ async def test_streaming_ingest_from_blob(self, is_managed_streaming):
                 self.streaming_ingest_client.ingest_from_blob(blob_descriptor, ingestion_properties)
 
         await self.assert_rows_added(2, timeout=120)
+
+    def test_json_401(self):
```
ohadbitt, line 569: can this be UT?

AsafMah (approved):

ohadbitt (approved):

Issue #567, opened 2025-02-05
Ingest from dataframe is broken for dataframes containing objects/dynamic columns - should ingest from JSON instead


drhar:
#### Problem description

I have an ADX table with dynamic columns that I want to store json objects in (e.g. arrays, dicts as per https://learn.microsoft.com/en-us/kusto/query/scalar-data-types/dynamic?view=microsoft-fabric. As these structures exist as objects in pandas dataframes I was naively using `ingest_from_dataframe` and didn't really notice at first that this was doing anything wrong as it passes the data type validation in kusto. However, arrays with length > 1 were actually being sent as an array containing a single string which was a load of single quoted items.

Now I've looked under the covers, it's clear that all this library does when ingesting from dataframe is dump the dataframe to CSV then ingest from that. Dumping to CSV ruins json formatting in pandas (single quotes strings). I feel like you should dump to json (ensure date format is compatible with ADX timestamps with `df.to_json(filename, orient="records", date_format="iso"`) instead and ingest from that file.

I'm now doing this instead as a workaround:

Old method (broken):

```python
        log.info("Ingesting data into Azure Data Cluster")
        ingestion_properties = IngestionProperties(
            database=DATABASE,
            table=TABLE,
            data_format=DataFormat.JSON,
        )
        rsp = ingest_client.ingest_from_dataframe(
            df, ingestion_properties=ingestion_properties
        )
```

New method:

```python
        df.to_json(filename, orient="records", date_format="iso"`)

        log.info("Ingesting data into Azure Data Cluster")
        ingestion_properties = IngestionProperties(
            database=DATABASE,
            table=TABLE,
            data_format=DataFormat.MULTIJSON,
        )
        rsp = ingest_client.ingest_from_file(
            filename, ingestion_properties=ingestion_properties
        )
```

#### Output of ``pip freeze`` 

<details>

aiohappyeyeballs==2.4.4
aiohttp==3.11.11
aiosignal==1.3.2
anyio==4.8.0
appdirs==1.4.4
astroid==3.3.8
async-timeout==5.0.1
attrs==24.3.0
azure-core==1.32.0
azure-identity==1.19.0
azure-kusto-data==4.6.3
azure-kusto-ingest==4.6.3
azure-storage-blob==12.24.0
azure-storage-queue==12.12.0
bandit==1.8.2
black==24.10.0
certifi==2024.12.14
cffi==1.17.1
charset-normalizer==3.4.1
click==8.1.8
[...]

PR #568, opened 2025-02-18
Fixed bug where nested json inside pandas wouldn't be ingested correctly

AsafMah:
solves #567

### DIFF
--- a/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
@@ -107,7 +107,7 @@ class BaseIngestClient(metaclass=ABCMeta):
         https://docs.microsoft.com/en-us/azure/data-explorer/ingest-data-overview#ingestion-methods
         :param pandas.DataFrame df: input dataframe to ingest.
         :param azure.kusto.ingest.IngestionProperties ingestion_properties: Ingestion properties.
-        :param DataFormat data_format: Format to convert the dataframe to. If not specified, it will try to infer it from the mapping, if not found, it will default to JSON.
+        :param DataFormat data_format: Format to convert the dataframe to - Can be DataFormat.CSV, DataFormat.JSOn or None. If not specified, it will try to infer it from the mapping, if not found, it will default to JSON.
         """
 
         if self._is_closed:
@@ -126,8 +126,10 @@ class BaseIngestClient(metaclass=ABCMeta):
                 is_json = False
         elif data_format == DataFormat.CSV:
             is_json = False
-        elif data_format != DataFormat.JSON:
-            raise ValueError("Unsupported format: {}".format(data_format))
+        elif data_format == DataFormat.JSON:
+            is_json = True
+        else:
+            raise ValueError("Unsupported format: {}. Supported formats are: CSV, JSON, None".format(data_format))
 
         file_name = "df_{id}_{timestamp}_{uid}.{ext}.gz".format(id=id(df), timestamp=int(time.time()), uid=uuid.uuid4(), ext="json" if is_json else "csv")
         temp_file_path = os.path.join(tempfile.gettempdir(), file_name)
--- a/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
@@ -121,11 +121,7 @@ class BaseIngestClient(metaclass=ABCMeta):
         is_json = True
 
         # If we are given CSV mapping, or the mapping format is explicitly set to CSV, we should use CSV
-        if (
-            not data_format
-            and ingestion_properties is not None
-            and (ingestion_properties.ingestion_mapping_type == DataFormat.CSV)
-        ):
+        if not data_format and ingestion_properties is not None and (ingestion_properties.ingestion_mapping_type == DataFormat.CSV):
             is_json = False
         elif data_format == DataFormat.CSV:
             is_json = False
--- a/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
+++ b/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
@@ -120,11 +120,11 @@ class BaseIngestClient(metaclass=ABCMeta):
 
         is_json = True
 
-        # If we are given CSV mapping, or the format is explicitly set to CSV, we should use CSV
+        # If we are given CSV mapping, or the mapping format is explicitly set to CSV, we should use CSV
         if (
             not data_format
             and ingestion_properties is not None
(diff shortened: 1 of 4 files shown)


azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
```diff
@@ -117,12 +115,11 @@ def ingest_from_dataframe(self, df: "pandas.DataFrame", ingestion_properties: In
...
-        file_name = "df_{id}_{timestamp}_{uid}.csv.gz".format(id=id(df), timestamp=int(time.time()), uid=uuid.uuid4())
+        file_name = "df_{id}_{timestamp}_{uid}.json.gz".format(id=id(df), timestamp=int(time.time()), uid=uuid.uuid4())
         temp_file_path = os.path.join(tempfile.gettempdir(), file_name)
-
-        df.to_csv(temp_file_path, index=False, encoding="utf-8", header=False, compression="gzip")
-
-        ingestion_properties.format = DataFormat.CSV
+        with gzip.open(temp_file_path, "wb") as temp_file:
+            df.to_json(temp_file, orient="records", date_format="iso", lines=True)
+        ingestion_properties.format = DataFormat.JSON
```
yogilad, line 122: thought JSON was obsolete.
multijson?

azure-kusto-ingest/tests/test_e2e_ingest.py
```diff
@@ -529,15 +529,20 @@ async def test_streaming_ingest_from_dataframe(self):
             "xtextWithNulls",
             "xdynamicWithNulls",
```
yogilad, line 530: nit: but column name is not correct anymore (not sure if the table is created ad-hoc or not)

azure-kusto-ingest/tests/test_e2e_ingest.py
```diff
@@ -529,15 +529,20 @@ async def test_streaming_ingest_from_dataframe(self):
...
-        rows = [
-            [0, "00000000-0000-0000-0001-020304050607", 0.0, 0.0, 0, 0, 0, 0, 0, 0, 0, 0, "2014-01-01T01:01:01Z", "Zero", "Zero", "0", "00:00:00", None, ""]
-        ]
+
+        guid = uuid.uuid4()
+
+        dynamic_value = ["me@dummy.com", "you@dummy.com", "them@dummy.com"]
+        rows = [[0, str(guid), 0.0, 0.0, 0, 0, 0, 0, 0, 0, 0, 0, "2014-01-01T01:01:01Z", "Zero", "Zero", "0", "00:00:00", None, dynamic_value]]
         df = DataFrame(data=rows, columns=fields)
         ingestion_properties = IngestionProperties(database=self.test_db, table=self.test_table, flush_immediately=True, data_format=DataFormat.CSV)
```
yogilad, line 538: Leftover CSV

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
```diff
@@ -117,12 +115,11 @@ def ingest_from_dataframe(self, df: "pandas.DataFrame", ingestion_properties: In
         if not isinstance(df, DataFrame):
             raise ValueError("Expected DataFrame instance, found {}".format(type(df)))
 
-        file_name = "df_{id}_{timestamp}_{uid}.csv.gz".format(id=id(df), timestamp=int(time.time()), uid=uuid.uuid4())
+        file_name = "df_{id}_{timestamp}_{uid}.json.gz".format(id=id(df), timestamp=int(time.time()), uid=uuid.uuid4())
```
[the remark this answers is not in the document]
yogilad: Not there's another hidden braking change here.
CSV matching is commonly ordinal while json matching is by name.
This means that if the dataframe names do not match the table names the mapping will fail.

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
AsafMah: Is it? JSON means JSONLINES, which is why I pass lines=True.
It's more performant and smaller

azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py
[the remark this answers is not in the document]
AsafMah: Discussed offline

yogilad (dismissed):

ViaFerrata:
Could it be that this has created a bug @AsafMah (tested with latest pypi release)?
I think the data format is not inferred correctly from the IngestionProperties now.

My Pandas ingestion from before the patch looked like this:
```python
ingestion_props = IngestionProperties(database=kusto_db, table=table_name, data_format=DataFormat.CSV, ignore_first_record=True, report_level=ReportLevel.FailuresAndSuccesses)
result_pd = client_qi.ingest_from_dataframe(df_csv, ingestion_properties=ingestion_props)
print(result_pd)
```

Output (before patch):
```
IngestionResult(status=IngestionStatus.QUEUED, database=log, table=ssh_log_data, source_id=xxx, obfuscated_blob_uri=https://xxx/yyy.csv.gz)
```

With the same code after the patch, json is chosen even though csv is explicitely selected within the IngestionProperties.
Output after patch:
```
IngestionResult(status=IngestionStatus.QUEUED, database=log, table=ssh_log_data, source_id=xxx, obfuscated_blob_uri=https://xxxx/yyy.json.gz)
```

Only noticed this, because the ingestion now fails since `json` and `ignore_first_record=True` is not compatible.

yogilad:
@ViaFerrata , 

IIUC, our impl. of the Pandas ingestion does not generate a header record.
It sounds like you were losing the first record of each ingestion.
Can you confirm?

@AsafMah , we should set this value to False in all ingestion paths (JSON and CSV)

ViaFerrata:
> @ViaFerrata ,
> 
> IIUC, our impl. of the Pandas ingestion does not generate a header record. It sounds like you were losing the first record of each ingestion. Can you confirm?

Looking at the source code (don't have access to my code right now), I think so, yes (due to `header=False`):
https://github.com/Azure/azure-kusto-python/blob/18d85ce8bbe9ea1f497f4ff78def2cc2a1cb518a/azure-kusto-ingest/azure/kusto/ingest/base_ingest_client.py#L141

I remember that I was fiddling around with the various ingestion options and reused the same IngestionProperties due to being lazy. Wanted to use the proper IngestionProperties for the pandas ingestion in the end, but seems like I forgot, thanks!

----

And regarding the data format, that is as intended, right (digged a bit into the source code now)? 
-> `data_format` in `IngestionProperties` is not related to the json/csv handover to `ingest_from_file`? Only the `ingestion_mapping_kind` plays a role.

Was just a bit confused since a month ago above code would handover csv files to `ingest_from_file` instead of json (according to the blob_uri).

PR #574, opened 2025-05-04
fix dependency issues in setup.py

AsafMah:

### DIFF
--- a/azure-kusto-ingest/setup.py
+++ b/azure-kusto-ingest/setup.py
@@ -46,8 +46,11 @@ setup(
     package_data={"": ["py.typed"]},
     install_requires=[
         "azure-kusto-data=={}".format(VERSION),
-        "azure-storage-blob==12.24.0",  # TODO - this has to be locked to this version due to https://github.com/Azure/azure-sdk-for-python/issues/40041
-        "azure-storage-queue>=12,<13",
+
+        # TODO - this has to be locked to this version due to https://github.com/Azure/azure-sdk-for-python/issues/40041
+        "azure-storage-blob==12.24.0",  
+        "azure-storage-queue==12.12.0",
+        
         "tenacity>=8.0.0",
     ],
     extras_require={"pandas": ["pandas"], "aio": []},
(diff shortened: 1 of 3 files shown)


azure-kusto-ingest/setup.py
```diff
@@ -45,6 +44,11 @@
     ],
     packages=find_packages(exclude=["azure", "*tests*", "*tests.*"]),
     package_data={"": ["py.typed"]},
-    install_requires=["azure-kusto-data=={}".format(VERSION), "azure-storage-blob>=12,<13", "azure-storage-queue>=12,<13", "tenacity>=8.0.0"],
+    install_requires=[
+        "azure-kusto-data=={}".format(VERSION),
+        "azure-storage-blob==12.24.0",  # TODO - this has to be locked to this version due to https://github.com/Azure/azure-sdk-for-python/issues/40041
+        "azure-storage-queue>=12,<13",
```
ohadbitt, line 50: queue should be the same

ohadbitt (approved):

azure-kusto-ingest/setup.py
```diff
@@ -45,6 +44,12 @@
     ],
     packages=find_packages(exclude=["azure", "*tests*", "*tests.*"]),
     package_data={"": ["py.typed"]},
-    install_requires=["azure-kusto-data=={}".format(VERSION), "azure-storage-blob>=12,<13", "azure-storage-queue>=12,<13", "tenacity>=8.0.0"],
+    install_requires=[
+        "azure-kusto-data=={}".format(VERSION),
+        # TODO - this has to be locked to this version due to https://github.com/Azure/azure-sdk-for-python/issues/40041
+        "azure-storage-blob==12.24.0",
```
ohadbitt, line 50: actually i see now they had issues with 12.24.0

ohadbitt (changes requested):

azure-kusto-ingest/setup.py
AsafMah: Fixed already

PR #577, opened 2025-05-15
annotate ClientRequestProperties

danield137:
add field annotations

### Changed
* added annotations to `client_request_id`, `application`, and `user` fields in `ClientRequestProperties` class


AsafMah:
Note that we're using black to format, and a specific version of it to be consistent.
There's a back_to_black.bat to run it.

For now it's just a newline so I'll remove it

AsafMah:
If you don't like the formatting we can also type them in the ctor

AsafMah (approved):

PR #646, opened 2026-08-12
Allow additional Application Insights tables in E2E tests

komal-rani522:
Updates the synchronous and asynchronous Application Insights E2E tests to require the expected baseline tables while allowing additional service-managed tables.


azure-kusto-data/tests/test_e2e_data.py
```diff
@@ -301,15 +301,13 @@ def test_log_analytics_query(self):
...
-            for primary in result.primary_results:
-                counter += 1
-                for row in primary.rows:
-                    assert row["TableName"] == expected_table_name.__next__()
+            actual_table_names = {row["TableName"] for primary in result.primary_results for row in primary.rows}
+            expected_table_names = set(self.application_insights_tables())
+            missing_table_names = expected_table_names - actual_table_names
 
-            assert counter == 1
+            assert not missing_table_names, f"Missing expected Application Insights tables: {sorted(missing_table_names)}"
```
monishkadas-ms, line 310: Can we flip these two asserts? len(result.primary_results) == 1 is the structural check. If the response comes back with zero primary results (auth failure, wrong DB etc), actual_table_names is an empty set and this fails with "Missing expected Application Insights tables: [all 11 names]", which points the reader at a non-existent service regression instead of "we got no result table back". Asserting the shape first gives the accurate failure.

azure-kusto-data/tests/test_e2e_data.py
```diff
@@ -319,15 +317,13 @@ async def test_log_analytics_query_async(self):
...
```
monishkadas-ms, line 326: Same as above.

tanmaya-panda1 (approved):

monishkadas-ms (approved):

